PR #5, opened 2023-02-21
feat: Initial public API of at_chops in pure C

xavierchanth:
This PR is in the draft stage because it is still a work in progress.

However, I would look some feedback on the design patterns I've chosen thus far.


xavierchanth:
Some very naive assumptions have been made here, it is likely that the public API will change to use pointers to pass various structs around in the implementation.

include/at_chops_type.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <stddef.h>  // size_t
+#include <stdint.h>  // uint8_t
+#include <stdbool.h> // bool
+
+namespace DartType
```
cpswan: This feels like we're leaking an implementation detail from one SDK to another.

People using our C/C++ SDK(s) shouldn't know or have to care that the atServer is implemented in Dart.

In many ways I'd be more comfortable if things went the other way (after all [C Isn't A Programming Language Anymore](https://faultlore.com/blah/c-isnt-a-language/)), but that shouldn't be necessary - we should be able to define our types without any reference to implementation details elsewhere.

xavierchanth: Makes sense, I did start with the Dart implementation first, with the intent to slowly cull it into a more C-like interface.

The two big distinctions between this and the Dart interface (beyond the fact that there are no classes):
1. Is that the strategy pattern will be implemented as function pointers in a struct instead of object inheritance (mainly used for the encryption algorithms)
2. I've opted to pass the metadata structs into the decrypt/encrypt/sign/verifySignature functions in AtChops to avoid the need for several overloads, since the method parameters in Dart are the same as the contents of the Metadata anyways

xavierchanth:
This is a much simpler interface, which should be vastly more efficient than what I had before

JeremyTubongbanua:
closing, as #7 is further ahead

PR #7, opened 2023-05-18
feat: at_chops - base64 and AES 256 CTR 

JeremyTubongbanua:
## What I did

- Branched off of branch in #5 
- Made directories: `include/at_chops/` and `include/at_client/` fore more `.h` files.
- aes ctr implementation using MbedTLS
- base64 implementation using MbedTLS
- small edits to the CommandHandler python tool
- small edits to the tests inside the root CMake


include/at_chops/aes_ctr.h
```diff
@@ -0,0 +1,37 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <string.h>
+
+typedef struct {
+    unsigned char *res; // result of the encryption/decryption
+    size_t reslen; // length of the result written
```
xavierchanth, line 21: Usually the size / len variable comes first in the struct when you have variable length pointers in a struct, that way you can compute where things are located in the memory.

include/at_chops/aes_ctr.h
```diff
@@ -0,0 +1,37 @@
...
+} AESResult;
+
+/**
+ * @brief AES 256 CTR encrypt plaintext
+ * 
+ * @param key_base64 the base64 encoded AES 256 key (e.g. "1DPU9OP3CYvamnVBMwGgL7fm8yB1klAap0Uc5Z9R79g=")
+ * @param plaintext the plain text to encryt, must be null terminated `\0`
+ * @return AESResult* the result of the encryption
+ */
+AESResult *atchops_aes256ctr_encrypt(const char *key_base64, const unsigned char *plaintext);
```
xavierchanth: Encrypt and Decrypt are both missing a way of handling errors.

Typically this is done with an int return value, with zero for success and errors representing other values. The result would instead be passed into the function as a pointer, which also allows the end developer to choose their own buffer size.

An alternative which could be done here is to add an int to your AESResult, which would indicate success / error

xavierchanth: Also, there are still declarations in at_chops.h as well, please remove the ones which aren't being used (I assume the other ones aren't)

src/at_chops/aes_ctr.c
```diff
@@ -0,0 +1,173 @@
...
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <mbedtls/aes.h>
+#include "aes_ctr.h"
+#include "base64.h"
+
+#define AES_KEY_BITS 256
```
xavierchanth, line 14: How difficult is it to retrieve the key length from the key passed into the function? If we can do that, then we can support CTR for any length out of the box.

.vscode/c_cpp_properties.json
```diff
@@ -8,7 +8,10 @@
                 "${workspaceFolder}/include",
```
xavierchanth: Also, given that some of the included files are now nested in include/at_chops, does this need to change? Or does vscode already recursively check nested folders?

JeremyTubongbanua: Good point, will investigate.

include/at_chops/aes_ctr.h
JeremyTubongbanua: Done in [036fa72](https://github.com/atsign-foundation/at_c/pull/7/commits/036fa7202f72c3989ce7014d94ef70b3c8398f80)

JeremyTubongbanua: Interesting, I did not know this. !

src/at_chops/aes_ctr.c
JeremyTubongbanua: Done in https://github.com/atsign-foundation/at_c/pull/7/commits/b09d205d4afb7c078c8af7d1f3a898ffceb57e58

.vscode/c_cpp_properties.json
JeremyTubongbanua: https://github.com/atsign-foundation/at_c/pull/7/commits/9d868a803b2bf589bcf17137e55c16a9a0c29137 done

include/at_chops/aes_ctr.h
JeremyTubongbanua: Done in https://github.com/atsign-foundation/at_c/pull/7/commits/332615e1651372707d046cfd109c4ce6d893aa19

include/at_chops/aes_ctr.h
```diff
@@ -0,0 +1,46 @@
...
+typedef enum {
+    AES_128 = 128,
+    AES_192 = 192,
+    AES_256 = 256,
+} AESKeySize;
+
+typedef struct {
+    size_t reslen; // length of the result written
+    unsigned char *res; // result of the encryption/decryption
+    int status; // status code of the operation
```
xavierchanth: For the same reason, you probably want status as first not last, so that status is at 0, reslen is at sizeof(int). Instead of reslen at 0, status at reslen * sizeof(unsigned char) which is added complexity

src/at_chops/aes_ctr.c
xavierchanth: But key size can be derived from the key, can't it?

JeremyTubongbanua: The keys are encoded in Base64, so that would probably involve decoding it first which is I think an unnecessary operation for now.

xavierchanth: Noted

include/at_chops/aes_ctr.h
JeremyTubongbanua: https://github.com/atsign-foundation/at_c/pull/7/commits/fe41978068847ca1fb850acf39b189df297aa9e3

xavierchanth (approved):

PR #18, opened 2023-06-29
feat: add Verbose Logging Utility

### LINKED ISSUE #11: Verbose Logging Utility
I have a lot of printf's and printx's everywhere.

Maybe the logger would have some functions like:

```c
int atlogger_log(const char *title, const char *message); // assume null terminated strings
int atlogger_logx(const char *title, const const unsigned char *bytes, const size_t byteslen); // print bytes e.g. "00 01 02 03 bd"
```

Usage example

```c
int main()
{
    atlogger_log("RSA", "My message"); // output "RSA | My message:"
    return 0;
}
```

Also provide an enum "INFO, WARNING, SEVERE" that replaces char *title with the enum?

**EDIT NEW**
@HamdaanAliQuatil would there be a way to make it so I could do something like `start_logging(Enum.INFO)` at the start of my code to only allow "INFO" logs and above to display?

HamdaanAliQuatil:
Closes #11 

**- What I did**

Add Verbose Logger utility for at_c

**- How I did it**

- [x] Created header file `at_logger.h` under `include/at_client/`
- [x] Created `at_logger.c` under `src/at_client/` and added `atlogger_log` and `atlogger_logx` methods

Add Verbose Logger utility for at_c

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
 	url = https://github.com/espressif/esp-idf.git
 [submodule "deps/Arduino-CMake-Toolchain"]
 	path = deps/Arduino-CMake-Toolchain
-	url = https://github.com/a9183756-gh/Arduino-CMake-Toolchain
+	url = https://github.com/a9183756-gh/Arduino-CMake-Toolchain.git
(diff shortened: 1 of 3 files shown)


JeremyTubongbanua:
I added the git submodules that I think you accidentally deleted @HamdaanAliQuatil

JeremyTubongbanua (approved):
Feel free to merge if you think it is ready @HamdaanAliQuatil

PR #28, opened 2023-07-31
feat: repl

JeremyTubongbanua:
**- What I did**

- refactors to sha, aes, connection
- made `repl.c` as an example of how to piece everything together

**- How I did it**

- see changes

**- How to verify it**

- see changes

**- Description for the changelog**

I will plan to add more refactors and optimizations in future PRs

PR #29, opened 2023-08-07
feat: complete refactor, ci compilation + test, CMake configs

### LINKED ISSUE #21: Automated Tests
## What needs to be done

You can run the tests by using the python tool:

```sh
python tool.py -p desktop -f mbedtls build
python tool.py -p desktop -f mbedtls test
```

Then check that all of the lines has "passed" in it.

Example:

```sh
jeremytubongbanua@Jeremys-M2-Air at_c % python tool.py -p desktop -f mbedtls test 

Running tests...
test_rsapublicpopulate passed!
test_connection passed!
test_base64 passed!
test_rsaprivatepopulate passed!
test_rsaencrypt passed!
test_rsadecrypt passed!
test_rsasign passed!
test_aes_ctr passed!

All tests passed!
```

### LINKED ISSUE #23: Update CMake Minimum Required Version to 3.19
Make CMake Minimum Required version to 3.19 for "adequate" Apple Silicon support

https://stackoverflow.com/questions/71886569/cmake-on-mac-with-arm-m1-is-running-linker-with-x86-64-architecture-instead-of-a#:~:text=1%20Answer&text=For%20anyone%20running%20into%20the,for%20Apple%20Silicon%20is%203.19.

### LINKED ISSUE #30: Use `cmake --build build` instead of `make`
https://github.com/pr0g/cmake-examples/blob/main/installing/README.md

image

Found a nice resource on why we should use `cmake --build build` instead of `cd build && make`

TLDR; every system uses different tools (like Make or Ninja).. cmake provides one way to do all

JeremyTubongbanua:
**- What I did**

- completely remade `packages/atclient/CMakeLists.txt`
- can be consumed by other CMake packages via `find_package(atclient REQUIRED CONFIG)` in another CMake project
- tests now use CTest instead of running each binary individually through the python tool
- shifting over to a pure CMake solution instead of using the python tool as different OS's require different generators and flags
- CMake version minimum is now 3.19 as that is "adequate" for Apple Silicon. See #23 
- Encouraging  use of `cmake --build build --target install` instead of `make install` because we want to specify the generator with the `-G Ninja` or `-G "Unix Makefiles"` flags on the CMake configure step, so that we can target different generators. SEe #30 
- Automated tests using GitHub actions
- Removed `deps/` folder and git submodule tactic entirely.. use `FetchContent` Cmake module instead to fetch outside packages. The command also gets from releases (MbedTLS v3.4.1) so that we are not working directly from their trunk in case of any breaking changes they accidentally push. Summary: WE NOW depend on their RELEASE instead of their TRUNK branch.

**- How I did it**

- Completely commented out your CMakeLists.txt @XavierChanth sorry...

**- How to verify it**

- See passing unit tests

closes #30 
closes #28 
closes #21 
closes #23


tools/uninstall-atclient.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+sudo rm -f $(cat ../packages/atchops/build/install_manifest.txt)
```
xavierchanth: Careful about using relative paths in your bash scripts, these will be based on the shell and differ based on pwd. You can use the following to fix that:

```sh
FULL_PATH_TO_SCRIPT="$(realpath "${BASH_SOURCE[0]}")"
SCRIPT_DIRECTORY="$(dirname "$FULL_PATH_TO_SCRIPT")"
```

Then you can orient yourself based on the current script's dir by prefixing your paths with `"$SCRIPT_DIRECTORY"`

JeremyTubongbanua: Thank

JeremyTubongbanua: That script no longer exists

TylerTrott (dismissed):

JeremyTubongbanua:
Would like to merge @XavierChanth 

Wrote up the windows atchops edge case in a ticket: #31 

All is documented in the README as well

xavierchanth (dismissed):

xavierchanth:
Nice work!

JeremyTubongbanua:
Merge whenever plz and ty @XavierChanth

xavierchanth (approved):

PR #50, opened 2023-11-15
fix: parse CA certs from headers file

### LINKED ISSUE #49: at_c: ROOT_CERT should not be hardcoded
### Is your feature request related to a problem? Please describe.

At this moment, one of the root server certificates is hardcoded in the `connection.h` file:
https://github.com/atsign-foundation/at_c/blob/14d2107ec6f3ed6f6d3e60a59f8489a13dffa11e/packages/atclient/include/atclient/connection.h#L9-L39

It's not so much a security problem (since they are public certificates from the root server), but rather an issue that can impede functionality at a given moment (certificates expire, and this approach requires us to keep that part of the code up to date or a `root-cert.pem` file containing the updated certificate).

### Describe the solution you'd like

One option is to request the certificate chain to the root server, validate the certificates and use them to start the connection as we currently do.

Another option is to study if it's possible to avoid handling certificates in our code (SSL libraries usually have a way to do all this process automatically).

### Describe alternatives you've considered

OpenSSL: While it may seem simpler and has good support, I don't think the change is worth it if the entire connection system with MbedTLS is already in place. Additionally, it may have adverse implications for embedded systems that I'm not familiar with (I'm just noting this as an option in case other complications unrelated to this issue arise).

### Additional context

_No response_

realvarx:
**- What I did**
Created a file named `cacerts.h` containing a collection of some of the CA certificates used by our servers.

**- How I did it**
When a connection is established, a certificate chain is created from the variables defined in `cacerts.h`, which will be used to verify and confirm the authenticity of the certificates sent to us by the servers to which connections are being made. This helps to prevent potential spoofing/mitm attacks.

Reading certificates directly from a file has not been possible due to a problem related to how different compiled programs interpret a variable that defines a path relative to their location. This problem only left us with the solution of locating this file in a known path (eg: `~/.atsign/at_c/`, which could cause problems for embedded systems), or downloading the certificates from the internet and storing them locally (which could also be a problem).

**- How to verify it**
Try the `pkam_authenticate` example or print information about the certificate chain by adding the following code after the `mbedtls_x509_crt_parse` method call in the `connection.c`file:
```
mbedtls_x509_crt *crt = &(ctx->cacert);
while(crt != NULL) {
    char buf[1024];
    mbedtls_x509_crt_info(buf, sizeof(buf) - 1, "", crt);
    printf("%s\n", buf);

    crt = crt->next;
}
```

Closes #49


cpswan:
@realvarx I've set this to draft whilst you figure out the issue with the tests. Set it back when you're ready for review.

cpswan (approved):

JeremyTubongbanua (approved):

cpswan:
@realvarx I left this one for you to merge. Is there still something else to be done?

realvarx:
> @realvarx I left this one for you to merge. Is there still something else to be done?

No, everything is fine. I just forgot to merge it

PR #64, opened 2023-12-22
feat: Notifications impl

realvarx:
## What I did

- Added uuid4 and cjson libraries
- Modified `atclient_init` to use new `load_keys` and `copy_atkeys` functions (now atkeys load using cjson)
- Created an at_talk sender demo (which will evolve to a full-communication demo in the future). 
- Created `notify` and `attalk_send` functions 
- Created part of `get_encryption_key_shared_by_me` and `get_encryption_key_shared_by_other` functions (except for the case on which the key needs to be created and stored).
- Created atsign struct and init/free functions
- Created `atutils.c` and `atutils.h` for several functions I found useful or necessary
- More minor changes/tweaks which aren't explicitly mentioned (note: many lines appear as modified when in reality they are not, it is simply due to the VSCode formatter introducing additional spaces in conditionals and function calls)

## WORK IN PROGRESS⚠️
- Manage memory releasing in most of the new methods
- Monitor implementation
- Metadata processing (from/to json/string)
- AtEvent and AtEventQueue structs and methods for event management

## Issues
 - There is a problem where the .h files do not detect the libraries, however this does not happen with the .c files; I must be doing something wrong in the configuration of the CMakeLists.txt file. This can cause the tests to fail or it may not be possible to run any of the demos. You can avoid the error by commenting out everything related to AtEvent (this is the header file that does not have visibility of cjson).


JeremyTubongbanua:
Hi Alvaro, I'll attempt to setup the CMake to recognize uuidv4 and cjson based on your work. Look forward to a PR sometime this week

xavierchanth:
superseded by #234

PR #70, opened 2024-01-17
feat: CMake dynamic library fetching + install/clean_install scripts

JeremyTubongbanua:
**- What I did**

- Building atclient and/or atchops will now dynamically fetch libraries. there is no longer a need to specify flags like `-DATCLIENT_FETCH_MBEDTLS=ON`. The CMake script will check if you already have it installed, otherwise, it will use FetchContent to fetch the source and build & install it it for your project.
- atclient and atchops each have `install.sh` and `clean_install.sh` for installing the library in your `/usr/local` directory (or similar)
- Removed macos-latest from CI because ubuntu and macos -latest both use x86_64 architecture and the same compiler and so I don't think it would make a difference

**- How I did it**

- See changes

**- How to verify it**

- Tests passing
- Now have to call only `cmake -S . -B build` instead of the old `cmake -S . -B build -DATCLIENT_FETCH_MBEDTLS=ON -DATCLIENT_FETCH_ATCHOPS=ON`


packages/atchops/tools/clean_install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -eu
+cd ..
+sudo rm -rf build
```
xavierchanth: This is a bit dangerous, I always try to do this relative to the script in case someone accidentally runs this file from another directory

xavierchanth: example:
```sh
FULL_PATH_TO_SCRIPT="$(realpath "${BASH_SOURCE[0]}")"
SCRIPT_DIRECTORY="$(dirname "$FULL_PATH_TO_SCRIPT")"
```

xavierchanth: /bin/bash compliant

packages/atchops/tools/run_ctest.sh
```diff
@@ -2,7 +2,7 @@
 set -eu
 cd ..
 sudo rm -rf build/tests
```
xavierchanth: Same thing here

packages/atclient/tools/clean_install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -eu
+cd ..
+sudo rm -rf build
```
xavierchanth: and here

xavierchanth (dismissed):

JeremyTubongbanua:
@XavierChanth 
Is this okay?

https://github.com/atsign-foundation/at_c/pull/70/commits/38fc03734787fc78219902baf6b09a21893dd065

realvarx (approved):

PR #101, opened 2024-02-05
feat: extended atclient unit test timeout, fixed atlogger headers, stringutils + unit tests, atclient_atkey_from_string + unit tests

### LINKED ISSUE #76: at_c: atclient_atkey_from_string
This ticket is blocked by #78 , #79 , and #80 . Those 3 tickets should be completed before this.

Implement taking an AtKey from a string and populating an `atclient_atkey` struct

A good example can be found here:
https://github.com/atsign-foundation/at_java/blob/trunk/at_client/src/main/java/org/atsign/client/util/KeyStringUtil.java

Located in `atclient/src/atkey.c`
```c
int atclient_atkey_from_string(atclient_atkey *atkey, const char *atkeystr, const unsigned long atkeylen)
{
    // 6 scenarios:
    // 1. PublicKey:            "public:name.wavi@smoothalligator"
    //      name == "name"
    //      sharedby = "smoothalligator"
    //      sharedwith = NULL
    //      namespace = "wavi"
    //      type = "public"
    //      cached = false
    // 2. PublicKey (cached):   "cached:public:name.wavi@smoothalligator"
    //      name == "name"
    //      sharedby = "smoothalligator"
    //      sharedwith = NULL
    //      namespace = "wavi"
    //      type = "public"
    //      cached = true
    // 3. SharedKey:            "@foo:name.wavi@bar"
    //      name == "name"
    //      sharedby = "bar"
    //      sharedwith = "foo"
    //      namespace = "wavi"
    //      cached = false
    // 4. SharedKey (cached):   "cached:@bar:name.wavi@foo"
    //      name == "name"
    //      sharedby = "foo"
    //      sharedwith = "bar"
    //      namespace = "wavi"
    //      cached = true
    // 5. PrivateHiddenKey:     "_latestnotificationid.wavi@smoothalligator"
    //      name == "_latestnotificationid"
    //      sharedby = "smoothalligator"
    //      namespace = "wavi"
    //      cached = false
    // 6. SelfKey:              "name.wavi@smoothalligator"
    //      name == "name"
    //      sharedby = "smoothalligator"
    //      sharedwith = NULL
    //      namespace = "wavi"
    //      cached = false
    // more scenarios
    // 7. No Namespace, SelfKey:
    // "name@smoothalligator"
    //      name == "name"
    //      sharedby = "smoothalligator"
    //      sharedwith = NULL
    //      namespace = NULL
    //      cached = false

    return 1; // TODO: implement
}
```

Also include unit tests

JeremyTubongbanua:
closes #76 

**- What I did**

- Set atclient unit tests timeout from 2 seconds to 10 seconds
- Fixed header resolution from atclient/atlogger.h to atlogger/atlogger.h
- Wrote stringutils.h and stringutils.c with unit tests
- Wrote implementation for `int atclient_atkey_from_string` and wrote 12 unit tests for it

**- How I did it**

- New two unit test files passing

**- How to verify it**

New test that tests all of the functions

stringutils.h
```c
int atclient_stringutils_trim_whitespace(const char *string, const unsigned long stringlen, char *out, const unsigned long outlen, unsigned long *outolen);

int atclient_stringutils_starts_with(const char *string, const unsigned long stringlen, const char *prefix, const unsigned long prefixlen);

int atclient_stringutils_ends_with(const char *string, const unsigned long stringlen, const char *suffix, const unsigned long suffixlen);
```

test_atkey_from_string.c
```c
// Test 1: public keys
// 1A: cached public key
#define TEST_ATKEY_FROM_STRING_1A "cached:public:publickey@bob"
// 1B: non-cached public key
#define TEST_ATKEY_FROM_STRING_1B "public:publickey@alice"
// 1C. non-cached public key with namespace
#define TEST_ATKEY_FROM_STRING_1C "public:name.wavi@jeremy"
// 1D. cached public key with namespace
#define TEST_ATKEY_FROM_STRING_1D "cached:public:name.wavi@jeremy"
// Test 2: shared keys
// 2A: non-cached shared key with namespace
#define TEST_ATKEY_FROM_STRING_2A "@alice:name.wavi@bob"
// 2B: cached shared key without namespace
#define TEST_ATKEY_FROM_STRING_2B "cached:@bob:name@alice"
// 2C: non-cached shared key without namespace
#define TEST_ATKEY_FROM_STRING_2C "@bob:name@alice"
// 2D: cached shared key with namespace
#define TEST_ATKEY_FROM_STRING_2D "cached:@bob:name.wavi@alice"
// Test 3: private hidden keys
// 3A: private hidden key
#define TEST_ATKEY_FROM_STRING_3A "_lastnotificationid@alice123_4😘"
// Test 4: self keys
// 4A: self key with no namespace
#define TEST_ATKEY_FROM_STRING_4A "name@alice"
// 4B: self key with namespace
#define TEST_ATKEY_FROM_STRING_4B "name.wavi@jeremy_0"
```


JeremyTubongbanua:
Ready for review

packages/atclient/src/stringutils.c
```diff
@@ -0,0 +1,156 @@
...
+int atclient_stringutils_starts_with(const char *string, const unsigned long stringlen, const char *prefix)
+{
+    int ret = -1;
+    if (string == NULL || prefix == NULL)
+    {
+        ret = -1;
+        goto exit;
+    }
+
+    if (stringlen == 0)
```
xavierchanth: I think you also need to check if string is shorter than prefix

xavierchanth: Otherwise strncmp will try to check beyond string's length

packages/atclient/src/stringutils.c
```diff
@@ -0,0 +1,156 @@
...
+int atclient_stringutils_ends_with(const char *string, const unsigned long stringlen, const char *suffix)
+{
+    int ret = -1;
+    if(string == NULL || suffix == NULL)
+    {
+        ret = -1;
+        goto exit;
+    }
+    if(strlen == 0)
+    {
```
xavierchanth, line 103: Same thing here, string length vs suffix length

packages/atclient/include/atclient/stringutils.h
```diff
@@ -0,0 +1,50 @@
...
+ * @param string the string to split, this will not be modified
+ * @param stringlen the length of the string (use strlen(string) if it is null-terminated)
+ * @param delim the delimiter to split the string by
+ * @param tokens the array of strings to write to
+ * @param tokensarrlen the length of the tokens array (maximum number of tokens)
+ * @param tokensolen the numbers of tokens actually used in the array
+ * @param tokenlen the string max length of each individual token
+ * @return int 0 on success, non-zero on failure
+ */
+int atclient_stringutils_split(const char *string, const unsigned long stringlen, const char *delim, char **tokens, const unsigned long tokensarrlen, unsigned long *tokensolen, const unsigned long tokenlen);
```
xavierchanth: To be consistent with trim_whitespace, can you swap tokensolen and tokenlen or outlen and outolen.

packages/atclient/tests/test_stringutils.c
```diff
@@ -0,0 +1,174 @@
...
+    memset(out, 0, sizeof(char) * outlen);
+    unsigned long outolen = 0;
+
+    const unsigned long tokenlen = 4096;
+    const unsigned long tokenslen = 16;
+    char **tokens = malloc(sizeof(char *) * tokenslen);
+    memset(tokens, 0, sizeof(char *) * tokenslen);
+    for(int i = 0; i < tokenslen; i++)
+    {
+        *(tokens + i) = malloc(sizeof(char) * tokenlen);
```
xavierchanth: This is a very expensive split operation, since we are allocating the full length of the string for each split... instead, we should do this allocation inside the loop of the split function.

xavierchanth (changes requested):
Need to be careful of memory allocation for the split function, if there are 10 splits needed, then we allocate 10x the memory necessary. Ideally we do the full memory allocation inside, since we also don't know how many tokens to allocate either.

xavierchanth (dismissed):

xavierchanth:
I'm going to approve this because it can be optimized later, but please leave some TODOs somewhere in the code

packages/atclient/include/atclient/stringutils.h
JeremyTubongbanua: This was on purpose because they're grouped like that:

The 3 are together (because they are related to the tokens array of strings) **"tokens"**
`char **tokens, const unsigned long tokensarrlen, unsigned long *tokensolen`

Then this 1 is solo (because it is related to the individual token strings) **"token"**
`const unsigned long tokenlen`

packages/atclient/src/stringutils.c
JeremyTubongbanua: Thank you good catch!
https://github.com/atsign-foundation/at_c/pull/101/commits/c0a8cbb369c8c06a9d332d4c87b82bfbc765341c

packages/atclient/src/stringutils.c
JeremyTubongbanua: Thank you, good catch
https://github.com/atsign-foundation/at_c/pull/101/commits/c0a8cbb369c8c06a9d332d4c87b82bfbc765341c

packages/atclient/tests/test_stringutils.c
JeremyTubongbanua: I can see what you are saying here.

Something I'm having trouble thinking about here is if we malloc the individual tokens of varying length ourselves, then we will have to free it ourselves as well. Then, we will have to provide the developer a free function too.

Then at that point, we may as well create our own array functionality which is going to turn into a huge mess.

Let me know what you think

cpswan:
@XavierChanth makes a bunch of good points, but I'm also wondering why we're writing this from scratch when I'd expect this sort of thing to be library code?

packages/atclient/tests/test_stringutils.c
xavierchanth: I mean, you can malloc (current size + (num splits - 1 bytes)) then copy each split, inserting a null terminator in between. Freeing the memory should be relatively easy, iterate over the array and free each pointer.

xavierchanth: Actually, since we discard the delimiter, we can simply replace the delimiter with a null terminator and call it a day. There will be allocated memory in between the null terminator and the start of the next element if the delimiter is more than a one character, but that's fine.

xavierchanth:
> @XavierChanth makes a bunch of good points, but I'm also wondering why we're writing this from scratch when I'd expect this sort of thing to be library code?

I've noticed that with these kinds of things, it's very easy to find code examples, and much harder to find good libraries that do just the one or few things we need.

packages/atclient/tests/test_stringutils.c
realvarx: `realloc` can also be expensive but maybe works fine in this context to keep it simple (same pointer, no delimiters...)

cpswan (approved):

PR #106, opened 2024-02-12
feat: atkey to string

### LINKED ISSUE #77: at_c: atclient_atkey_to_string
Implement taking a populated atclient_atkey struct and converting it into a string such as `public:publickey@bob`

See: https://github.com/atsign-foundation/at_java/blob/trunk/at_client/src/main/java/org/atsign/common/Keys.java#L46
for an example

Located in `atclient/src/atkey.c`

```c
int atclient_atkey_to_string(atclient_atkey atkey, char *atkeystr, unsigned long *atkeystrlen, unsigned long atkeystrolen)
{
    return 1; // TODO: implement
}
```

JeremyTubongbanua:
closes #77 

**- What I did**
- implemented `atclient_atkey_to_string`
- Wrote unit tests for this newly implemented function

*other stuff i did*
- CMakeLists.txt format
- removed private hidden key as per @gkc 's advice, and made according fixes

**- How to verify it**
- `test_atkey_to_string.c`

### DIFF
--- a/packages/atclient/src/atkey.c
+++ b/packages/atclient/src/atkey.c
@@ -11,8 +11,6 @@
 
 #define TAG "atkey"
 
-#define ATKEY_GENERAL_BUFFER_SIZE 4096 // sufficient memory for keyName, namespace, sharedWith, and sharedBy strings
-
 void atclient_atkey_init(atclient_atkey *atkey)
 {
     memset(atkey, 0, sizeof(atclient_atkey));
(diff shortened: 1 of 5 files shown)


packages/atclient/src/atkey.c
```diff
@@ -221,9 +217,125 @@ int atclient_atkey_from_atstr(atclient_atkey *atkey, const atclient_atstr atstr)
...
+        ret = atclient_atstr_append(&string, "public:");
+        if(ret != 0)
+        {
+            atclient_atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "atclient_atstr_append_literal failed\n");
+            goto exit;
+        }
+    }
+    else if(atkey.metadata.ispublic == 1 || atkey.atkeytype == ATCLIENT_ATKEY_TYPE_PUBLICKEY)
+    {
+        atclient_atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_WARN, "either atkey's metadata ispublic is not set to 1 or atkey's atkeytype is not set to ATCLIENT_ATKEY_TYPE_PUBLICKEY for atkey: %.*s\n", (int) atkey.name.olen, atkey.name.str);
```
xavierchanth, line 248: Should we just set metadata.ispublic to 1 if the key type is ATCLIENT_ATKEY_TYPE_PUBLICKEY

xavierchanth: Seems like we should trust the key type in the code to be correct, over some metadata which is abstracted underneath it.

xavierchanth: Oh wait, I guess we wouldn't set it, but we should probably assume that the key is in fact a public key. We can warn, but still proceed with appending "public:"

xavierchanth (approved):

xavierchanth:
Looks good @JeremyTubongbanua

packages/atclient/src/atkey.c
JeremyTubongbanua: I was thinking we could have some sort of "fix atkey" function that fixes stuff like this

xavierchanth: > I was thinking we could have some sort of "fix atkey" function that fixes stuff like this

Yeah that would be a good parallel to the Atkey builder pattern we use in the dart sdk

PR #109, opened 2024-02-13
style: Adopt a clang format config

xavierchanth:
**- What I did**

- I was originally trying to keep the style guidelines from changing too significantly, while trying to keep true to our typical Dart / Google style guidelines.
- I still ended up with a ton of formatting changes in every file...
- So I settled on LLVM with line length 120 to keep the rules simple.

**- How I did it**

**- How to verify it**

style: Adopt a clang format config

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 ---
 Language: Cpp # Covers C and C++
 BasedOnStyle: LLVM # Added this to ensure a sane set of defaults
-BreakBeforeBraces: Allman # atsign-foundation/at_c is already heavily styled like this, so lets keep this setting as is
 ColumnLimit: 120 # LLVM uses 80, but we mostly use 120 at Atsign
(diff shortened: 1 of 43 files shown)


xavierchanth:
This is a really big diff, but it is just straight up formatting (and the .clang-format file). If we want to go with a Microsoft base instead I can raise another PR.

gkc (approved):

cpswan (approved):

realvarx (approved):

JeremyTubongbanua:
Is there any documentation (or tool) @XavierChanth so I (and other developers) can adopt this style and remain consistent?

xavierchanth:
> Is there any documentation (or tool) @XavierChanth so I (and other developers) can adopt this style and remain consistent?

Yes, the default settings in vscode are "file" - fallback to "Visual Studio":
CleanShot 2024-02-14 at 12 53 28@2x

Now that we have a .clang-format file, vscode should just pick up the new formatting rules.

xavierchanth:
Also @JeremyTubongbanua `clang-format -i \<path>` for example:

`clang-format -i src/*.c` and `clang-format -i include/**/*.h` in the context of our repos

Normally clang-format, like sed, will output the changes to stdout. `-i` tells clang-format to do it in-place.

PR #111, opened 2024-02-14
feat: atclient_atsign & get_enc_shared_keys

### LINKED ISSUE #107: at_c: atsign type
A typedef struct holding an atsign that ensures correct format

To be used in functions that use atsigns (for example, atkey).

realvarx:
**- What I did**
- Impl of `atclient_atsign` following https://github.com/atsign-foundation/at_c/issues/107 discussion (new methods: `atclient_atsign_init` and `atclient_atsign_free`)
- Refactored `atclient_get_encryption_key_shared_by_me` and `atclient_get_encryption_key_shared_by_other` methods
- `atclient_connection_disconnect`
- Minor tweaks to other methods/files

**- How to verify it**
- Run work-in-progress `at_talk` demo 

Closes https://github.com/atsign-foundation/at_c/issues/107


packages/atclient/3rdparty/uuid4/include/uuid4.h
JeremyTubongbanua, line 1: Did you mean to commit 3rdparty/uuid4 ?

We use FetchContent to get uuid4 from our own [fork](https://github.com/atsign-foundation/uuid4)

packages/atclient/include/atclient/atsign.h
```diff
@@ -1,14 +1,32 @@
...
+
+// Structure to represent the AtSign class
+typedef struct atclient_atsign {
+    char* atsign;
+    char* without_prefix_str;
+} atclient_atsign;
+
+// Function to initialize an AtSign object
+int atclient_atsign_init(atclient_atsign *atsign, char *atsign_str);
+
```
JeremyTubongbanua: Since we don't modify `char *atsign_str`, ideally this parameter would be `const char *atsign_str` to indicate we only plan on reading this argument.

packages/atclient/include/atclient/atclient.h
```diff
@@ -56,4 +61,10 @@ int atclient_get(atclient *ctx, const char *key, char *value, const unsigned lon
 int atclient_delete(atclient *ctx, const char *key);
 void atclient_free(atclient *ctx);
 
+int atclient_get_encryption_key_shared_by_me(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_me);
+int atclient_get_encryption_key_shared_by_other(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_other);
+// int atclient_attalk_send(atclient *ctx, atsign *recipient, char *enc_key_shared_by_me, char *msg);
```
JeremyTubongbanua: Since we're writing to `char *enc_key_shared_by_me`, we should also pass in the buffer length of this pointer and the output length. We would need the buffer length to do error checking, such as making sure we are not writing outside of the bounds of the buffer. Then, output length is useful so that the programmer knows the length of the output string.

What is your opinion on a function header such as :
```
int atclient_get_encryption_key_shared_by_me(const atclient ctx, const atclient_atsign recipient, char *enckeysharedbyme, const size_t enckeysharedbymelen, size_t *enckeysharedbymeolen);
```

JeremyTubongbanua (commented):
Left some of my thoughts. None of these are concrete and would like everyone's input as well.

packages/atclient/include/atclient/atclient.h
realvarx: Here the output length is always going to be the same (if everything goes fine, if not, then we are facing an error and we will probably exit the method anyway). The user already has the control of the buffer length (the user decides how much memory to allocate for that variable). If I change `strcpy` by `strncpy` we should be fine against buffer overflows for this specific case (because the expected output length is always the same).

packages/atclient/include/atclient/atsign.h
realvarx: Solved https://github.com/atsign-foundation/at_c/pull/111/commits/584d8d8f52f57716346dba973d9149b8292a378f

packages/atclient/include/atclient/atclient.h
JeremyTubongbanua: Oh, are base64 encoded aes keys always 44 characters in length or something?

.gitignore
```diff
@@ -20,3 +20,5 @@ __pycache__/
 .pio/
 
 test/*/test_*
+
+.clang-format
```
xavierchanth: .clang-format shouldn't be gitignored, it should be committed to the repo

xavierchanth: merge trunk into your branch to pick up the one I added

packages/atclient/3rdparty/uuid4/include/uuid4.h
realvarx: Further discussion in https://github.com/atsign-foundation/at_c/issues/112

packages/atclient/include/atclient/atclient.h
realvarx: I think so

.gitignore
realvarx: Yes sorry, I added some extra config while doing a quick test in my IDE and forgot to remove that line. Changed https://github.com/atsign-foundation/at_c/pull/111/commits/0bd52ddcd19705004910c079b5a19f8a7beeb281

packages/atclient/src/atclient.c
```diff
@@ -239,3 +245,144 @@ void atclient_free(atclient *ctx) {
...
+
+int atclient_get_encryption_key_shared_by_me(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_me)
+{
+    //  atclient_atkeys atkeys, const char *myatsign
+    int ret = 1;
+
+    // llookup:shared_key.recipient_atsign@myatsign
+    char *command_prefix = "llookup:shared_key.";
+
+    char *command = (char *)malloc(strlen(command_prefix) + (strlen(ctx->atsign.atsign)*2 - 1) + 3);
```
xavierchanth: This could be a `char command[strlen(command_prefix) + (strlen(ctx->atsign.atsign)*2 - 1) + 3];` instead, no need to allocate heap memory if the pointer doesn't need to live beyond this scope.

packages/atclient/src/atclient.c
```diff
@@ -239,3 +245,144 @@ void atclient_free(atclient *ctx) {
...
+    strcpy(command, command_prefix);
+    strcat(command, recipient->without_prefix_str);
+    strcat(command, ctx->atsign.atsign);
+    strcat(command, "\r\n");
+
+    const unsigned long recvlen = 1024;
+    unsigned char *recv = (unsigned char *)malloc(sizeof(unsigned char) * recvlen);
```
xavierchanth: Same thing here

packages/atclient/src/atclient.c
```diff
@@ -239,3 +245,144 @@ void atclient_free(atclient *ctx) {
   atclient_connection_free(&(ctx->root_connection));
   atclient_connection_free(&(ctx->secondary_connection));
 }
+
+int atclient_get_encryption_key_shared_by_me(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_me)
```
xavierchanth: I can't tell whats an input parameter and whats an output parameter, make sure to prefix input only pointers with `const <type>* <name>`

packages/atclient/src/atclient.c
```diff
@@ -239,3 +245,144 @@ void atclient_free(atclient *ctx) {
...
+        }
+        memcpy(enc_key_shared_by_me, plaintext, plaintextlen);
+    }
+    
+    else if (atclient_stringutils_starts_with(recv, recvlen, "error:AT0015-key not found", strlen("error:AT0015-key not found")))
+    {
+        // TODO: or do I need to create, store and share a new shared key?
+    }
+    return 0;
+}
```
xavierchanth, line 322: Both those pieces of memory I called out are never freed, nor is the ownership of the memory passed to something outside this scope

packages/atclient/src/atclient.c
```diff
@@ -239,3 +245,144 @@ void atclient_free(atclient *ctx) {
...
+            printf("atchops_rsa_decrypt (failed): %d\n", ret);
+            return ret;
+        }
+        memcpy(enc_key_shared_by_other, plaintext, plaintextlen);
+    }
+    else if (atclient_stringutils_starts_with(recv, recvlen, "error:AT0015-key not found", strlen("error:AT0015-key not found")))
+    {
+        // TODO: or do I need to create, store and share a new shared key?
+    }
+    return 0;
```
xavierchanth: Same thing

packages/atclient/src/atsign.c
```diff
@@ -1,13 +1,37 @@
...
+  if ((strlen(atsign_str) > 56) || (atsign_str == NULL) || (strlen(atsign_str) == 0)) {
+    ret = 1;
+    atclient_atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "atclient_atsign_init: %d\n", ret);
+    return ret;
+  }
+
+  memset(atsign, 0, sizeof(atclient_atsign));
+  atsign->atsign = malloc(strlen(atsign_str) + 1);
+
+  const unsigned long maxatlen = 56;
```
xavierchanth: I think this may need to be 57? 1 for @, 55 for name part, 1 for null term. Or is the `@` included in the 55 count?

xavierchanth: Can confirm, this is 1 "@" + 55 "\<name part>" + 1 '\0' for a total of 57 chars

xavierchanth: Okay, so apparently the maximum length is 55 utf7 characters, after the utf8 (single char) string is converted to utf7 for persistence

xavierchanth: Okay, so it turns out that this limit needs to be "@" + 55 **utf7** chars + null terminator. But, since this is enforced by the atSign length on the registrar, I think it is okay to use a buffer size slightly larger, since it is at least large enough to hold any production atsign at 55 utf8 chars

packages/atclient/include/atclient/atclient.h
```diff
@@ -56,4 +61,10 @@ int atclient_get(atclient *ctx, const char *key, char *value, const unsigned lon
 int atclient_delete(atclient *ctx, const char *key);
 void atclient_free(atclient *ctx);
 
+int atclient_get_encryption_key_shared_by_me(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_me);
+int atclient_get_encryption_key_shared_by_other(atclient *ctx, atclient_atsign *recipient, char *enc_key_shared_by_other);
```
JeremyTubongbanua: Missing documentation on header functions.

Important to mention which arguments are required, optional, also important to mention that *ctx is assumed to be initialized via the atclient_init function.

realvarx: https://github.com/atsign-foundation/at_c/pull/111/commits/3bebe594e2b76ba75d3e3649e536e91a9fa1e5be

packages/atclient/src/atclient.c
realvarx: https://github.com/atsign-foundation/at_c/pull/111/commits/1a1d78bdc2dd9df247f84e288b091ed51edba434

packages/atclient/src/atclient.c
realvarx: https://github.com/atsign-foundation/at_c/pull/111/commits/369643d0062b13cd8134954bbb9e060514d33f22

packages/atclient/src/atsign.c
realvarx: https://github.com/atsign-foundation/at_c/pull/111/commits/29c2134259a2e506a74eb6db84438f5c2f588e2d

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
   atclient_connection_free(&(ctx->root_connection));
   atclient_connection_free(&(ctx->secondary_connection));
 }
+
+static int atclient_create_shared_encryption_key(atclient *ctx, const atclient_atsign *recipient,
+                                                 char *enc_key_shared_by_me);
+
+static int atclient_get_public_encryption_key(atclient *ctx, const atclient_atsign *atsign,
```
xavierchanth, line 250: Make sure to put these in the header file, I had to move them in order to properly use these in sshnpd

examples/at_talk/src/main.c
```diff
@@ -0,0 +1,111 @@
...
+  atclient.atsign = myatsign;
+
+  // Init atkeysfile and read keys
+  atclient_atkeysfile atkeysfile;
+  atclient_atkeysfile_init(&atkeysfile);
+  ret = atclient_atkeysfile_read(&atkeysfile, ATKEYSFILE_PATH);
+  if (ret != 0) {
+    atclient_atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_INFO, "atclient_atkeys_file_read: %d\n", ret);
+    atclient_free(&atclient);
+    atclient_atkeysfile_free(&atkeysfile);
```
JeremyTubongbanua, line 43: Should use ATLOGGER_LOGGING_LEVEL_ERROR instead of INFO

JeremyTubongbanua (dismissed):

packages/atclient/src/atsign.c
```diff
@@ -1,13 +1,37 @@
...
+    return ret;
+  }
+
+  atsign->without_prefix_str = atsign->atsign + 1;
+  return ret;
+}
+
+void atclient_atsign_free(atclient_atsign *atsign) { free(atsign->atsign); }
+
 int atclient_atsign_without_at_symbol(char *atsign, const unsigned long atsignlen, unsigned long *atsignolen,
```
xavierchanth, line 35: do we still need this?

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
...
+int atclient_get_encryption_key_shared_by_me(atclient *ctx, const atclient_atsign *recipient,
+                                             char *enc_key_shared_by_me) {
+  int ret = 1;
+
+  // llookup:shared_key.recipient_atsign@myatsign
+  char *command_prefix = "llookup:shared_key.";
+  short command_prefix_len = 19;
+  short atsign_with_at_len = (short)strlen(ctx->atsign.atsign);
+
+  short command_len = command_prefix_len + (atsign_with_at_len * 2 - 1) + 3;
```
xavierchanth, line 262: shouldn't the length be `command_prefix_len + (recipient length) + atsign_with_at_len`;

xavierchanth: Simplified with (I usually leave these kinds of simplifications as a comment):
-1 to remove the recipient's `@` + 1 for null term

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
...
+      atclient_atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "atchops_rsa_decrypt: %d\n", ret);
+      return ret;
+    }
+
+    memcpy(enc_key_shared_by_me, plaintext, plaintextlen);
+  }
+
+  else if (atclient_stringutils_starts_with(recv, recvlen, "error:AT0015-key not found",
+                                            strlen("error:AT0015-key not found"))) {
+    // TODO: or do I need to create, store and share a new shared key?
```
xavierchanth, line 313: I don't think we need to create / share a new shared key, what if we just want to check whether a key has been shared or not.

We should have a special return code to allow us to identify this situtation.

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
...
+  char *command_prefix = "llookup:shared_key.";
+  short command_prefix_len = 19;
+  short atsign_with_at_len = (short)strlen(ctx->atsign.atsign);
+
+  short command_len = command_prefix_len + (atsign_with_at_len * 2 - 1) + 3;
+  char command[command_len];
+  snprintf(command, command_len, "llookup:shared_key.%s%s\r\n", recipient->without_prefix_str, ctx->atsign.atsign);
+
+  const unsigned long recvlen = 1024;
+  unsigned char recv[sizeof(unsigned char) * recvlen];
```
xavierchanth, line 267: this should be `unsigned char recv[recvlen]` - it's not like malloc. The number inside of the `[ ]` is the number of `unsigned char` allocated in the array.

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
...
+  memset(recv, 0, sizeof(unsigned char) * recvlen);
```
xavierchanth, line 268: Here, we do need to multiply by `sizeof(unsigned char)` so this line is correct.

packages/atclient/src/atclient.c
```diff
@@ -239,3 +243,300 @@ void atclient_free(atclient *ctx) {
...
+    response[strlen(response) - response_prefix_len - 1] = '\0';
+  }
+
+  // does my atSign already have the recipient's shared key?
+  if (atclient_stringutils_starts_with(response, recvlen, "data:", 5)) {
+    response = response + 5;
+
+    // 44 + 1
+    unsigned long plaintextlen = 45;
+    unsigned char plaintext[sizeof(unsigned char) * plaintextlen];
```
xavierchanth, line 297: Same thing here

xavierchanth: There are more in subsequent functions, I'm not going to call each one out.

xavierchanth (approved):

xavierchanth:
Let's merge this and fix these things in another PR. I need to start using some of these functions

packages/atclient/src/atsign.c
realvarx: Not right now, but could be useful. We can always remove it for now and put it back if we need it

packages/atclient/src/atclient.c
realvarx: This is what all our other SDKs do. I know C focus is different, but I think it's the way the method should be working. Is there any other reason (besides obtaining it or creating it if it does not exist) why a developer would want to know whether or not a key has already been created? We can introduce an extra opcode/bool parameter to let the user decide if there's a common situation on which it is helpful

xavierchanth: Yeah, we don't need things to be so automatic in the C sdk. We should focus on things being simple and relinquishing control to the application developer. It will help us keep things light weight and easier to maintain.

packages/atclient/src/atsign.c
realvarx: Ohh actually it is used in `atclient_pkam_authenticate`:
https://github.com/atsign-foundation/at_c/blob/18a4357e07887c4cc5a38cb78f665592290f0052/packages/atclient/src/atclient.c#L109

PR #118, opened 2024-02-20
fix: various tweaks to previous get_enc_shared_keys PR

realvarx:
**- What I did**

Various tweaks discussed in https://github.com/atsign-foundation/at_c/pull/111 and some extra

- [x] Put atclient_create_shared_encryption_key and atclient_get_public_encryption_key in header file
- [x] Should use ATLOGGER_LOGGING_LEVEL_ERROR instead of INFO in demo
- [x] Fix incorrect command_len
- [x] Shouldn't multiply by sizeof() inside []
- [x] Add docs to `atclient_create_shared_encryption_key` and `atclient_get_public_encryption_key`
- [x] Replace unsigned long by size_t in new atclient methods
- [x] Missing const in some variables
- [x] Add `create_new_if_not_found` bool param to `atclient_get_encryption_key_shared_by_me` and new error code


packages/atclient/src/atclient.c
```diff
@@ -365,10 +359,10 @@ int atclient_get_encryption_key_shared_by_other(atclient *ctx, const atclient_at
     response = response + 5;
 
     // 44 + 1
-    unsigned long plaintextlen = 45;
-    unsigned char plaintext[sizeof(unsigned char) * plaintextlen];
+    size_t plaintextlen = 45;
```
JeremyTubongbanua: Maybe use `const size_t` here?

realvarx: Sure, I can also mark as `const` all `command_prefix_len` variables

realvarx: https://github.com/atsign-foundation/at_c/pull/118/commits/56c34a767b50d40d2234ba39159909399b52c076

JeremyTubongbanua (approved):

PR #123, opened 2024-02-22
chore: remove externally available atstr functions 

JeremyTubongbanua:
**- What I did**
- Refactored atkey.c/.h and atkeys.c/.h so the atstr functions (that are not used anywhere) cannot be used by the outside developer.

### DIFF
--- a/packages/atclient/src/atkey.c
+++ b/packages/atclient/src/atkey.c
@@ -213,10 +213,6 @@ exit: {
 }
 }
 
-int atclient_atkey_from_atstr(atclient_atkey *atkey, const atclient_atstr atstr) {
-  return atclient_atkey_from_string(atkey, atstr.str, atstr.olen);
-}
-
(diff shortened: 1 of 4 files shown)


packages/atclient/src/atkey.c
```diff
@@ -312,20 +308,6 @@ exit: {
 }
 }
 
-int atclient_atkey_to_atstr(const atclient_atkey atkey, atclient_atstr *atstr) {
```
realvarx, line 315: If they are helpful for you (if we finally keep some methods using `atclient_atstr`), why not keeping them as `static`? They wont be visible outside this file.

Before usage in the same `.c` file:
`static int atclient_atkey_from_atstr(atclient_atkey *atkey, const atclient_atstr atstr);`

(same for other methods)

JeremyTubongbanua: They are not helpful. In fact, they are not used anywhere. I made them in case I needed them in the future

realvarx (approved):

PR #146, opened 2024-02-27
chore: refine types and memory

xavierchanth:
**- What I did**

- make things const where possible
- leave some TODOs with improvements to be made
- remove typedefs from some enums which shouldn't need to be stored (often) by the developer
  - e.g. aes/rsa key size enums
- removed unused buffer sizes and moved needed ones to their associated header file
  - kept constants.h, since it will be used by monitor/notify
- removed root connection from atclient so it can be used by multiple secondary connections
  - e.g. monitor thread only needs root to do pkam auth then it can release it back to the main connection
**- How I did it**

**- How to verify it**

chore: refine types and memory

### DIFF
--- a/packages/atchops/include/atchops/aes.h
+++ b/packages/atchops/include/atchops/aes.h
@@ -2,10 +2,11 @@
 #ifndef ATCHOPS_AES_H
 #define ATCHOPS_AES_H
 
-typedef enum atchops_aes_keysize {
+enum atchops_aes_size {
+  ATCHOPS_AES_NONE = 0,
   ATCHOPS_AES_128 = 128, // not tested
   ATCHOPS_AES_256 = 256,
-} atchops_aes_keysize;
+};
 
 /**
  * @brief Generate an AES key of size keylen bits
@@ -14,7 +15,7 @@ typedef enum atchops_aes_keysize {
  * @param keysize key length in bits (e.g. AES-256 = 256 => ATCHOPS_AES_256)
  * @return int 0 on success
  */
-int atchops_aes_generate_key(unsigned char *key, const atchops_aes_keysize keysize);
+int atchops_aes_generate_key(unsigned char *key, const enum atchops_aes_size keysize);
 
 /**
  * @brief Generate an AES key of size keylen bits encoded in base 64
@@ -26,6 +27,6 @@ int atchops_aes_generate_key(unsigned char *key, const atchops_aes_keysize keysi
  * @return int 0 on success
  */
 int atchops_aes_generate_keybase64(unsigned char *keybase64, const unsigned long keybase64len,
-                                   unsigned long *keybase64olen, atchops_aes_keysize keysize);
+                                   unsigned long *keybase64olen, const enum atchops_aes_size keysize);
 
 #endif
--- a/packages/atchops/include/atchops/rsa.h
+++ b/packages/atchops/include/atchops/rsa.h
@@ -4,6 +4,12 @@
 #include "atchops/rsakey.h"
 #include <mbedtls/md.h>
 
+enum atchops_rsa_size {
+  ATCHOPS_RSA_NONE = 0,
+  ATCHOPS_RSA_2048 = 2048,
+  ATCHOPS_RSA_4096 = 4096,
+};
+
 /**
  * @brief Sign a message with an RSA private key
  *
@@ -16,8 +22,8 @@
  * @param signatureolen the length of the signature buffer after signing
  * @return int 0 on success
  */
-int atchops_rsa_sign(atchops_rsakey_privatekey privatekey, mbedtls_md_type_t mdtype, const unsigned char *message,
-                     const unsigned long messagelen, unsigned char *signaturebase64,
+int atchops_rsa_sign(const atchops_rsakey_privatekey privatekey, const mbedtls_md_type_t mdtype,
+                     const unsigned char *message, const unsigned long messagelen, unsigned char *signaturebase64,
                      const unsigned long signaturebase64len, unsigned long *signaturebase64olen);
 
 /**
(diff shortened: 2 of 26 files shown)


packages/atchops/include/atchops/rsa.h
```diff
@@ -16,8 +22,8 @@
  * @param signatureolen the length of the signature buffer after signing
  * @return int 0 on success
  */
-int atchops_rsa_sign(atchops_rsakey_privatekey privatekey, mbedtls_md_type_t mdtype, const unsigned char *message,
-                     const unsigned long messagelen, unsigned char *signaturebase64,
+int atchops_rsa_sign(const atchops_rsakey_privatekey privatekey, const mbedtls_md_type_t mdtype,
```
JeremyTubongbanua, line 25: Should we use a const pointer instead of just a const object to avoid making a copy? 

`const atchops_rsakey_privatekey *privatekey`

atchops_rsakey_privatekey is a humungous struct

packages/atchops/include/atchops/aes.h
```diff
@@ -2,10 +2,11 @@
 #ifndef ATCHOPS_AES_H
 #define ATCHOPS_AES_H
 
-typedef enum atchops_aes_keysize {
+enum atchops_aes_size {
```
JeremyTubongbanua, line 5: I think it's cool that we're dropping the typedef, but I also don't see the justification for the change.

MbedTLS's hash enums are a typedef, which we take as an argument in rsa_sign. Maybe it would be good to remain consistent?

```c
typedef enum {
    MBEDTLS_MD_NONE=0,    /**< None. */
    MBEDTLS_MD_MD5=0x03,       /**< The MD5 message digest. */
    MBEDTLS_MD_RIPEMD160=0x04, /**< The RIPEMD-160 message digest. */
    MBEDTLS_MD_SHA1=0x05,      /**< The SHA-1 message digest. */
    MBEDTLS_MD_SHA224=0x08,    /**< The SHA-224 message digest. */
    MBEDTLS_MD_SHA256=0x09,    /**< The SHA-256 message digest. */
    MBEDTLS_MD_SHA384=0x0a,    /**< The SHA-384 message digest. */
    MBEDTLS_MD_SHA512=0x0b,    /**< The SHA-512 message digest. */
    MBEDTLS_MD_SHA3_224=0x10,  /**< The SHA3-224 message digest. */
    MBEDTLS_MD_SHA3_256=0x11,  /**< The SHA3-256 message digest. */
    MBEDTLS_MD_SHA3_384=0x12,  /**< The SHA3-384 message digest. */
    MBEDTLS_MD_SHA3_512=0x13,  /**< The SHA3-512 message digest. */
} mbedtls_md_type_t;
```

packages/atclient/include/atclient/atclient.h
```diff
@@ -69,13 +69,17 @@ int atclient_pkam_authenticate(atclient *ctx, const atclient_atkeys atkeys, cons
...
  * @param atclient the atclient context (must satisfy the two conditions stated above)
+ * @param root_conn initialized root connection
  * @param atkey the populated atkey to put the value into (must satisfy the two conditions stated above)
  * @param value the value to put into atServer
  * @param valuelen the length of the value (most of the time you will use strlen() on a null-terminated string for this
  * value)
  * @return int 0 on success
  */
-int atclient_put(const atclient *atclient, const atclient_atkey *atkey, const char *value, const size_t valuelen);
+int atclient_put(atclient *atclient, atclient_connection *root_conn, const atclient_atkey *atkey, const char *value,
```
JeremyTubongbanua, line 79: What would atclient_put need root_conn for?

JeremyTubongbanua: (Same for atclient_get_publickey, and atclient_get_sharedkey)

packages/atclient/include/atclient/atclient.h
```diff
@@ -110,14 +114,15 @@ int atclient_get_selfkey(const atclient *atclient, atclient_atkey *atkey, char *
...
  * @param atkey the populated atkey to get the value from (must satisfy the two conditions stated above)
  * @param value the buffer to hold value gotten from atServer
  * @param valuelen the buffer length allocated for the value
  * @param valueolen the output length of the value gotten from atServer
  * @return int 0 on success
  */
-int atclient_get_publickey(const atclient *atclient, const atclient_atkey *atkey, char *value, const size_t valuelen,
-                           size_t *valueolen);
+int atclient_get_publickey(atclient *atclient, atclient_connection *root_conn, const atclient_atkey *atkey, char *value,
+                           const size_t valuelen, size_t *valueolen);
```
JeremyTubongbanua, line 125: Is it true that you can't use the connections inside of atclient if it is passed in as a const pointer? I've been doing `get_selfkey` and `get_publickey` stuff just fine.

I thought that const pointer means it is pass by reference and you are not allowed to change the values inside (you can't say something like: `atclient->secondary_connection = some_new_connection()`)

JeremyTubongbanua (commented):
Thank you for these changes @XavierChanth

packages/atchops/include/atchops/aes.h
xavierchanth: I dropped the typedef because they really shouldn't be passing around the type and I want to reduce scope pollution where possible.

packages/atclient/include/atclient/atclient.h
xavierchanth: If we want / need to lookup the other atSign's encryption keys, feel free to change this

packages/atclient/include/atclient/atclient.h
xavierchanth: We shouldn't denote it as const because mbedtls_ssl_write requires a writeable ssl context.

packages/atchops/include/atchops/rsa.h
xavierchanth: This can be changed when someone goes to work on it.

xavierchanth:
@JeremyTubongbanua these are all nits, I need this merged before I travel tomorrow.

JeremyTubongbanua (approved):

PR #159, opened 2024-03-17
refactor: `unsigned long` -> `size_t`

### LINKED ISSUE #105: at_c: use size_t where unsigned long is and stdbool for booleans
Would like a discussion of making C99 our minimum C version (because stdbool is only supported in C99+). This would mean dropping support for ANSI C (C89/90).

The sprint points allocated to this ticket will consider two things:
1. going in and replacing `unsigned long` with `size_t` from the `stddef.h` header
2. converting from int to booleans (such as iscached and ispublic in metadata) using `bool` in `stdbool.h`
3. Update documentation to say that the minimum version is C99

Another interesting thing to note about using type headers is it guarantees the size of the type. Example: a int may be 8 bytes in some systems as opposed to the regular 4, but using something like `uint64_t` from `stdint.h` guarantees it to be 8 bytes all the time.

JeremyTubongbanua:
closes #105 

**- What I did**

- refactor: `unsigned long` -> `size_t`

**- How I did it**

- see changes

**- How to verify it**

- unit tests passing

### DIFF
--- a/packages/atchops/include/atchops/aes.h
+++ b/packages/atchops/include/atchops/aes.h
@@ -1,6 +1,7 @@
-
 #ifndef ATCHOPS_AES_H
 #define ATCHOPS_AES_H
+
+#include <stddef.h>
 
 enum atchops_aes_size {
   ATCHOPS_AES_NONE = 0,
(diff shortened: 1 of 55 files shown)


xavierchanth:
A related comment (no action needed):

If we ever do end up needing to support a legacy build system which doesn't support size_t / bool, we can use some build directives to conditionally add our own typedefs for size_t / bool.

realvarx (approved):

PR #167, opened 2024-03-20
chore: update repl to work

JeremyTubongbanua:
**- What I did**
- refactors took place, and broke the repl example, this PR fixes it
- also reduced buffer sizes to 4096

**- How to verify it**

image


realvarx (approved):
https://github.com/atsign-foundation/at_c/blob/5e11f24762e89f4dca18725f1106f4ae8c5e9dff/examples/desktop/repl/src/main.c#L102
Just a small suggestion, `@atsign command` or `@atsign@command` look more similar to our other repls or direct ssl client connection, but it's just a picky detail haha

xavierchanth (approved):

PR #181, opened 2024-04-05
fix: cjson header issue

### LINKED ISSUE #180: at_c: using current trunk as a library does not compile (cJSON header issue)

realvarx:
**- What I did**

- Fixed atclient `CMakeLists.txt` file so the `cJSON` library can be detected by other files. This involves using `FetchContent_Declare` to fetch the library, instead of using the files in 3rdparty folder.
- As a result of the previous changes, 3rdparty folder was removed, as it won't be necessary for now.

**- How to verify it**

Run any of the currently working examples in `examples/desktop/crud`.

Closes #180


JeremyTubongbanua (dismissed):

JeremyTubongbanua:
Thanks for this !

packages/atclient/CMakeLists.txt
```diff
@@ -139,18 +136,36 @@ if(NOT ESP_PLATFORM)
...
+      URL_HASH SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
+    )
 
-  # # ########################################################
+    list(APPEND ATCLIENT_TARGETS_TO_INSTALL cjson) # since we've fetched it, we have to install it
+    set(CJSON_BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
+    set(CJSON_OVERRIDE_BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+    set(ENABLE_CJSON_TEST OFF CACHE BOOL "" FORCE)
+    add_compile_options(-fPIC)
+    FetchContent_MakeAvailable(cJSON)
```
xavierchanth: this should be lower 'cjson' now right?

xavierchanth: 'cjson' is the correct package name

realvarx: Thanks, good catch https://github.com/atsign-foundation/at_c/pull/181/commits/eedece0d9339bdf397421108b3801343a5c849d0

packages/atclient/CMakeLists.txt
```diff
@@ -176,18 +191,25 @@ if(NOT ESP_PLATFORM)
   add_library(${PROJECT_NAME} STATIC ${atclient_srcs})
   add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
+  # Manually add header file
+  configure_file(${cjson_SOURCE_DIR}/cJSON.h ${CMAKE_BINARY_DIR}/include/cjson/cJSON.h)
+  
+  # Set include directories for atclient target
+  target_include_directories(${PROJECT_NAME} PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
+    $<BUILD_INTERFACE:${ATCLIENT_INCLUDE_DIR}>
```
JeremyTubongbanua, line 200: Is `${CMAKE_BINARY_DIR}/include` and `${ATCLIENT_INCLUDE_DIR}` the same thing?

packages/atclient/CMakeLists.txt
```diff
@@ -176,18 +191,25 @@ if(NOT ESP_PLATFORM)
...
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  )
+
+  # atclient depends on cjson, so the latter must be built first
+  add_dependencies(atclient cjson)
```
JeremyTubongbanua, line 205: Doesn't have to be done in this PR, but should we also consider adding `atchops`, `uuid4` and `mbedtls` as dependencies to atclient?

Is this the magic line of code that fixed everything btw?

JeremyTubongbanua (commented):
I had some questions

packages/atclient/CMakeLists.txt
realvarx: Apparently, they aren't depending on the host (I thought they were the same). That line has been the fix for GH Actions...

packages/atclient/CMakeLists.txt
realvarx: For some reason, when I changed to `FetchContent_Declare` `cjson` was building later than the other packages, so I had to add that extra line. It didnt fix the issue, but was part of the fix

packages/atclient/CMakeLists.txt
xavierchanth: I think `ATCLIENT_INCLUDE_DIR` points to `${<GNU INSTALL DIR>}/include/atclient`

packages/atclient/CMakeLists.txt
xavierchanth: I think we have a bunch of cleanup work to take care of in this file, but it can be dealt with later

xavierchanth (approved):

PR #186, opened 2024-04-10
feat: pkam refactor

### LINKED ISSUE #119: at_c: use atclient_atsign in atclient_pkam_authenticate
`atclient_pkam_authenticate` should be using new `atclient_atsign` (the one inside of `atclient` struct, or through a new parameter `atclient_atsign *recipient` which may be null, in which case the atsign will be the one inside of `atclient`) instead of `const char *atsign`:
https://github.com/atsign-foundation/at_c/blob/18a4357e07887c4cc5a38cb78f665592290f0052/packages/atclient/src/atclient.c#L60-L61

That will allow to stop using `atclient_atsign_without_at_symbol`:
https://github.com/atsign-foundation/at_c/blob/18a4357e07887c4cc5a38cb78f665592290f0052/packages/atclient/src/atclient.c#L109

### LINKED ISSUE #141: at_c: set atsign in pkam_authenticate
Once pkam_authenticate is successful,

`atclient.atsign` should be initialized to the atsign that pkam_authenticated.

### LINKED ISSUE #142: at_c: set atkeys on successful pkam_authenticate
Set the atkeys once pkam_authenticate is successful in the atclient struct

### LINKED ISSUE #161: at_c: optimize atclient_pkam_authenticate
optimize pkam_authenticate's memory footprint. currently it takes ~4KB of memory (see screenshot below) because of numerous buffer allocations that could instead be statically allocated.

realvarx:
**- What I did**

Refactored `atclient_pkam_authenticate`:
- Changed the way we build commands by using `snprintf`
- Changed the parameters of the method, so now ~~`atclient_atsign` is used and~~ both `atsign` and `atkeys` parameters are pointers
- ~~Removed `withoutat` variable related code, as it is no longer necessary~~
- Replaced `withoutat` variable by `atsign_without_prefix_str`
- Changed the calls to the method in the different desktop examples

**- How to verify it**

Run `pkam_authenticate` or any of the other desktop examples

image

Closes #119 
Closes #141 
Closes #142 
Closes #161


JeremyTubongbanua:
This pretty much closes #161 as well. I added it to your ticket.

Thanks for doing this :)

xavierchanth:
Will review on Monday

packages/atclient/include/atclient/atclient.h
```diff
@@ -55,8 +55,8 @@ int atclient_start_secondary_connection(atclient *ctx, const char *secondaryhost
  * @param atsign the atsign the atkeys belong to
  * @return int 0 on success
  */
-int atclient_pkam_authenticate(atclient *ctx, atclient_connection *root_conn, const atclient_atkeys atkeys,
-                               const char *atsign, const size_t atsignlen);
+int atclient_pkam_authenticate(atclient *ctx, atclient_connection *root_conn, const atclient_atkeys *atkeys,
+                               const atclient_atsign *atsign);
```
xavierchanth: I would prefer if we kept this as a char*, I find managing atclient_atsign to be very irritating, since you have to deep copy to clone. Where as with char* you can use strcpy directly. This also reduces the surface for bugs, and pushes to use an idiomatic C approach, (instead of `atsign.without_at` [atclient_atsign], call `atsign + 1` [char*])

xavierchanth (changes requested):
I don't think this change is necessary. When we originally designed the atclient_atsign struct, it seemed like an innocent change, but having learned more about how idiomatic C should be written, we should prefer the C native `char*` for the atsign type.

Having written monitor in a way that needs things like atkeys and atsigns be referenced across threads, it's clear we should prefer the char* for the sole reason of preventing confusion & bugs when the `atclient_atsign` struct is shallow cloned (and thus the underlying memory is the same).

xavierchanth:
Rather than pushing to use atclient_atsign everywhere, I think we should be doing the opposite, and pushing to eliminate its usage.

JeremyTubongbanua:
I agree with getting rid of atclient_atsign.. I haven't found it particularly useful except for ensuring that the atSign inputted into the function is correctly formatted, which could be done with a function check instead.

JeremyTubongbanua (dismissed):

JeremyTubongbanua:
It LGTM to merge though @realvarx

packages/atclient/include/atclient/atclient.h
realvarx: https://github.com/atsign-foundation/at_c/pull/186/commits/431a4c2c3493dd1f9808a2ee61f42ba8104c06b4

realvarx:
`atclient` struct currently has an `atclient_atsign` type inside, which some methods depend on. If both of you agree on eliminating `atclient_atsign` completely, I could try to also refactor that along the rest of the sdk in a different PR

JeremyTubongbanua (dismissed):

JeremyTubongbanua (approved):

xavierchanth (approved):

PR #196, opened 2024-04-19
chore: single atsdk export with a bunch of targets

xavierchanth:
**- What I did**

- Redo of the package management
- cmake/ folder contains the cmake code for the preferred way to fetch dependencies & build packages
  - Made it so we always use fetchcontent (no find_package) so that dependencies are coming from a consistent, pinned source & builds are reproducible (or as close as possible) across environments and runs
- added a root CMakeLists.txt which builds the atsdk project and exports all of our packages and dependencies:
  - atsdk::atclient
  - atsdk::atchops
  - atsdk::atlogger
  - atsdk::uuid4-static
  - atsdk::cjson
  - atsdk::mbedtls
  - atsdk::mbedcrypto
  - atsdk::mbedx509
  - atsdk::everest (transitive dependency of mbedtls)
  - atsdk::p256m (transitive dependency of mbedtls)
- Changed the "cjson/cJSON.h" to "cJSON.h" so that we can include straight from the build interface without directory manipulation

The ideal workflow is now:
- build atsdk using the root cmake
- use find_package or fetch content to link it in your application (as we do with our examples)

**- How I did it**

**- How to verify it**

chore: single atsdk export with a bunch of targets


cpswan (dismissed):
Tests need to be updated to work with this change. Essentially whatever we expect other folk to do when using the SDK, we need to do ourselves in the tests.

xavierchanth:
> Tests need to be updated to work with this change. Essentially whatever we expect other folk to do when using the SDK, we need to do ourselves in the tests.

The reason that tests fail is because most of us already have cjson installed on our local, so this enhancement I made fixes finding cjson based on the normal way of building.

GitHub actions runner doesn't have cjson installed, so when we try to build it, it seems that files are ending up in abnormal places. I'll have to redo the cjson build steps.


Will update this PR with a fix for Monday... cjson has been a thorn in our dependency toolchain...

xavierchanth:
@realvarx commit https://github.com/atsign-foundation/at_c/pull/196/commits/115c966bb8f1575f95e436b2964b09950bb756ad should disable install / exports on the desktop which are not needed when building for espidf.

xavierchanth:
Okay, this is properly ready for review.

JeremyTubongbanua (approved):

JeremyTubongbanua:
I can improve and start working on CI for using the SDK as a dependency

PR #208, opened 2024-04-27
refactor: make RSA and SHA functions more atomic

### LINKED ISSUE #182: at_c: atchops function parameters should take non-base64
This is the function header for AES CTR decrypt in atchops

`int atchops_aesctr_decrypt(const char *keybase64, const size_t keybase64len, const enum atchops_aes_size keybits,
                           unsigned char *iv, const unsigned char *ciphertextbase64,
                           const size_t ciphertextbase64len, unsigned char *plaintext,
                           const size_t plaintextlen, size_t *plaintextolen)`

In hindsight, this function shouldn't take any base64 input. In this case, it should take a raw `unsigned char *key` as well as raw `ciphertext`. Leave it up to the developer to base64 encode it for finer error handling and to make this function more atomic.

This also leaves room for optimization (5 subsequent AES decrypts in `atkeys.c` will be decoding the AES key 5 times when you really only need to decode the AES key once).

Task List:

```[tasklist]
### Tasks
- [x] AES CTR Encrypt/Decrypt functions
- [ ] RSA Encrypt/Decrypt
- [ ] RSA Sign/Verify
```

JeremyTubongbanua:
## - What I did
**- Made RSA encrypt/decrypt and sign/verify functions more atomic.** Before, we would expect a base64 input or would output a base64 string. This would result to unnecessary base64 operations that the developer may not have needed. For example, 5 back-to-back RSA decrypts would have 5 base64 decodes (since you would have to base64 decode the input) but what if the original ciphertext was already in raw bytes? Then That would have resulted in 5 unnecessary base64 decodes which is precious compute power. On the developer side, this would simply result in more control as to what is base64 encoded/decoded. The developer cost isn't that much either, as it is just one more step before the encryption/decryption operation, if they deem it necessary.
**- Removed unnecessary function arguments in SHA functions** (such as hashlen, which should always output to 32 depending on the SHA algorithm). In this case, we leave it up to the developer to properly allocate 32 bytes in their output buffer (as specified in the function header documentation). Previously, we would simply receive the output buffer size and error check that the output (32) was less than or equal to their buffer size (e.g. 1024). But in this case, the developer should be reading the docs to see that the output is ALWAYS related to their hashing algorithm size (e.g. SHA 256 has output of 32 bytes). this is also specified in the function header documentation. Hence, we do not need to output the length of the hash in the function, for example. This pattern is seen elsewhere in places like RSA sign and verify.

**- How I did it**
New rsa.h:
```c
/**
 * @brief Sign a message with an RSA private key
 *
 * @param privatekey the private key struct to use for signing, see atchops_rsakey_populate_privatekey
 * @param mdtype the hash type to use, see atchops_md_type, e.g. ATCHOPS_MD_SHA256
 * @param message the message to sign
 * @param messagelen the length of the message, most people use strlen() to find this length
 * @param signature the signature buffer to populate, must be pre-allocated. Signature size will correspond to the
 * specified hashing algorithm (e.g., this function expects `signature` to be a buffer of 32 bytes allocated, if using
 * ATCHOPS_MD_SHA256 (256 bits = 32 bytes)) .
 * @return int 0 on success
 */
int atchops_rsa_sign(const atchops_rsakey_privatekey privatekey, const mbedtls_md_type_t mdtype,
                     const unsigned char *message, const size_t messagelen, unsigned char *signature);

/**
 * @brief Verify a signature with an RSA public key
 *
 * @param publickey the public key to use for verification, see atchops_rsakey_populate_publickey
 * @param mdtype the hash type to use, see atchops_md_type, e.g. ATCHOPS_MD_SHA256
 * @param message the original message to hash, in bytes
 * @param messagelen the length of the original message, most people use strlen() to find this length
 * @param signature the signature to verify, expected to be the same length as the key size (e.g. 256 bytes for 2048 RSA modulus)
 * @return int 0 on success
 */
int atchops_rsa_verify(const atchops_rsakey_publickey publickey, const mbedtls_md_type_t mdtype, const unsigned char *message,
                       const size_t messagelen, unsigned char *signature);

/**
 * @brief Encrypt bytes with an RSA public key
 *
 * @param publickeystruct the public key struct to use for encryption, see atchops_rsakey_populate_publickey
 * @param plaintext the plaintext to encrypt, in bytes
 * @param plaintextlen the length of the plaintext, most people use strlen() to find this length
 * @param ciphertext the ciphertext buffer to populate
 * @param ciphertextsize the length of the ciphertext buffer
 * @param ciphertextlen the written length of the ciphertext buffer after encryption operation has completed, should be the same as the key size (e.g. 256 bytes for 2048 RSA modulus)
 * @return int 0 on success
 */
int atchops_rsa_encrypt(const atchops_rsakey_publickey publickey, const unsigned char *plaintext,
                        const size_t plaintextlen, unsigned char *ciphertext, const size_t ciphertextsize,
                        size_t *ciphertextlen);

/**
 * @brief Decrypt bytes with an RSA private key
 *
 * @param privatekey the private key struct to use for decryption, see atchops_rsakey_populate_privatekey
 * @param ciphertext the ciphertext to decrypt, in bytes
 * @param ciphertextlen the length of the ciphertext, should be the same as the key size (e.g. 256 bytes for 2048 RSA modulus)
 * @param plaintext the plaintext buffer to populate
 * @param plaintextsize the size of the plaintext buffer (allocated size)
 * @param plaintextlen the written length of the plaintext buffer after decryption operation has completed
 * @return int 0 on success
 */
int atchops_rsa_decrypt(const atchops_rsakey_privatekey privatekey, const unsigned char *ciphertext,
                        const size_t ciphertextlen, unsigned char *plaintext, const size_t plaintextsize,
                        size_t *plaintextlen);
```

new sha.h
```c
/**
 * @brief SHA Hash an input buffer
 *
 * @param mdtype the type of hash to use (e.g. MBEDTLS_MD_SHA256)
 * @param input the input to hash (in raw bytes)
 * @param inputlen the length of the input buffer (most likely strlen(input))
 * @param output (out) the output buffer to write the hash to. The length of this buffer should correspond to the hash
 * type (e.g. 32 bytes for SHA256 (256 bits = 32 bytes))
 * @return int 0 on success
 */
int atchops_sha_hash(const mbedtls_md_type_t mdtype, const unsigned char *input, const size_t inputlen,
                     unsigned char *output);

```

**- How to verify it**
- tests passing
- ran crud tests and they still pass

Closes #182


JeremyTubongbanua:
Requesting review @realvarx @XavierChanth 

Finished merge to trunk and did manual tests on crud operations

xavierchanth (approved):

PR #250, opened 2024-05-10
feat: added async_read bool to atclient

xavierchanth:
**- What I did**
- added async_read command to atclient
- notify, put, noop all return as success if the command sends, they don't synchronously listen for a response
- The idea is to enable this for the monitor thread to avoid race conditions (by reading from two threads) when you do notify / put / noop from the connection.
- One should never call get from the monitor thread since there is no way to identify which get command the asynchronous incoming value is from (there could be a use-case, but it is unlikely, thus unimplemented)
**- How I did it**

**- How to verify it**

feat: added async_read bool to atclient


xavierchanth:
I suspect I will be making a few more changes to this, thus I've set it to draft until I've made all the changes I need

xavierchanth:
put & notify need more thought since they get the public encryption key...

delete is relatively straightforward.

I have made note that async_read is experimental. Anyone who uses it is on their own

JeremyTubongbanua:
I think every CRUD & event operation that you can call should have its own `_async` variant (like atclient_put_sync, atclient_put_async, etc,.). That way, the caller has more ability to choose where the program should wait for the operation to fully go through.

My question is:

When you do `mbedtls_ssl_write`, doesn't that fill the `mbedtls_ssl_read` with the response? And wouldn't that infect the next time you run `mbedtls_ssl_write` and wanted to `mbedtls_ssl_read`? For example:

```c
mbedtls_ssl_write("notify:<...>"); // we sent a notification, we don't care about the id

// ... time goes by ...

mbedtls_ssl_write("llookup:<..>"); // we're getting data, we care about the response
mbedtls_ssl_read(buffer); // reading...

// Won't `buffer` have something like `notification:{...}\ndata:abcdefgh==` ? because the previous mbedtls_ssl_write had a response that was never read or something like that
```

JeremyTubongbanua (dismissed):

JeremyTubongbanua:
Good to merge for now as long as it works (none of the tests are failing)

But I think will need some more thinking down the line

packages/atclient/include/atclient/atclient.h
```diff
@@ -16,6 +16,9 @@ typedef struct atclient {
   atclient_connection secondary_connection;
   atclient_atsign atsign;
   atclient_atkeys atkeys;
+  // Warning! async_read is an experimental feature and not fully implemented.
+  // You should leave this set to false unless you know what you are doing.
```
JeremyTubongbanua, line 20: I think there should be more documentation on what async_read actually does

packages/atclient/src/atclient.c
```diff
@@ -259,7 +234,9 @@ int atclient_send_heartbeat(atclient *heartbeat_conn, bool listen_for_ack) {
   goto exit;
 
 exit: {
-  free(recv);
+  if (!heartbeat_conn->async_read) {
+    free(recv);
+  }
```
JeremyTubongbanua, line 239: I think we should avoid doing if checks on free-ing memory as much as possible to avoid reducing code complexity. free(NULL); is fine to do and recv is always = NULL unless pointed to memory location

JeremyTubongbanua: Because if we do this for a bunch of memory pointers, there'll be a gajillion if-checks and make it harder to read and track code

JeremyTubongbanua (approved):
Upon merge, I'll write a ticket to write some functional tests to ensure that everything works even with async_read on

JeremyTubongbanua:
With async_read set to true, we allow `delete` and `noop` to be ran, but...

a nice feature to have would be to enable specific scenarios to pass in async reads such as:

We don't need to do any reading when doing
- putting a publickey (inherently non-encrypted)
- putting a sharedkey that holds a non-encrypted value
- notifying that holds a non-encrypted value

PR #258, opened 2024-05-15
fix: notify commands that are >4096 bytes

JeremyTubongbanua:
**- What I did**
notify fixes & refactors
- Fixed a bug where notify commands that were greater than 4096 bytes would not execute. This is because the cmdbuffer was a constant size of 4096, instead of dynamically evaluating how much memory to allocate for the cmd string. This PR fixes that.
- applied base64 buffer optimizations
- New helper functions (for atclient_notify function)
- New functional test on atclient_notify for a long notiifcation (8012 bytes)

```h
static int generate_cmd(const atclient_notify_params *params, const char *cmdvalue, const size_t cmdvaluelen,
                        char **ocmd, size_t *ocmdolen);
static size_t calculate_cmd_size(const atclient_notify_params *params, const size_t cmdvaluelen, size_t *atkeyolen,
                                 unsigned long long *ottln, size_t *medatastrolen);
```

NEW base64 stuff

- New base64 functions (for calculating optimal base64 length)
```h
size_t atchops_base64_encoded_size(const size_t plaintextsize);
size_t atchops_base64_decoded_size(const size_t encodedsize);
```
- base64 tests on various encode/decode lengths

Chore stuff:
- Removed sudo from at_talk script

**- How I did it**

**- How to verify it**
- Added test for a VERY long notification
- Added base64 tests to test optimal buffer length functions
- Manual test at_talk as of this [commit](https://github.com/atsign-foundation/at_c/pull/258/commits/870be0c05d738999cfbbffc66269b29facfa1917) 

fix: notify command >4096 bytes


packages/atchops/src/base64.c
```diff
@@ -2,12 +2,22 @@
...
   return mbedtls_base64_decode(dst, dstsize, dstlen, src, srclen);
 }
+
+size_t atchops_base64_encoded_size(const size_t plaintextsize) {
+  return MAX((size_t)(plaintextsize * 1.5 + 0.5), 8); // add 0.5 to ceiling
+}
+
+size_t atchops_base64_decoded_size(const size_t encodedsize) {
+  return MAX((size_t)(encodedsize * 0.75 + 0.5), 8); // add 0.5 to ceiling
+}
```
xavierchanth: I think this needs to be `* 3 / 2` and `* 3 / 4` to benefit from the compiler optimizations that we discussed earlier, but there are more caveats related to endianness and how to efficiently address the ceiling portion.

It should be possible to perform both of these functions entirely as bit wise operations. But it is too much of a time sink now, this is something we should revisit.

Just so I remember for later:

Let p be the plaintext size.
Let x be the first bit from the little end of P (do an AND against 0b00000001).
Let y be p bit shifted towards the little end (essentially removing the x bit from p, also ensuring that the new bit is zeroed)

Then x + y + p should be the optimized version of `plaintextsize * 1.5 + 0.5`

---
A similar algorithm can be created for the decoded_size function, but computations must be done twice to essentially get a half and a quarter of the encoded size to add together. A bit wise OR of the two `x` bits from this algorithm can be added to the final result to perform the ceiling.

xavierchanth: Again, this is complete overkill, so for now, maybe change to:
`* 3 /2 + 1` and `*3/4+1` we may have a single bit of extra overhead, but that one bit will not hurt.

xavierchanth:
```suggestion
size_t atchops_base64_encoded_size(const size_t plaintextsize) {
  return MAX((size_t)(plaintextsize * 3 / 2 + 1), 8); // add 0.5 to ceiling
}

size_t atchops_base64_decoded_size(const size_t encodedsize) {
  return MAX((size_t)(encodedsize * 3 / 4 + 1), 8); // add 0.5 to ceiling
}
```

xavierchanth (approved):

PR #260, opened 2024-05-15
chore: moved fix_stdout_buffer to atlogger

JeremyTubongbanua:
**- What I did**
- We had duplicate code
- So I just moved it to atlogger so it could be used globally

**- How I did it**

**- How to verify it**

chore: moved fix_stdout_buffer to atlogger


xavierchanth (approved):
Great idea!

PR #273, opened 2024-05-22
chore: atchops maintenance

xavierchanth:
**- What I did**

- Extracted the mbedtls_md_type_t from our public functions and replaced it with an atchops_md_type enum. 

Note: There is a constant lookup map called atchops_mbedtls_md_map, this does use runtime memory which we can optimize away later by replacing with build-time macros, but what I've opted for right now is good enough. We accomplish the main goal which is to ensure that the public API won't break if we need to change crypto libraries in the future.

- Added atlogger to the atchops build and logging errors in rsa sign.

**- How I did it**

**- How to verify it**

- Lots of whitespace changes, I suggest reviewing with hide whitespace turned on

chore: atchops maintenance


CMakeLists.txt
```diff
@@ -26,13 +26,13 @@ endif()
 message(STATUS "[ATSDK] ATSDK_AS_SUBPROJECT: ${ATSDK_AS_SUBPROJECT}")
 
 # install each package
+message(STATUS "Building atlogger")
+add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/packages/atlogger)
+
```
xavierchanth, line 31: Had to move this to before atchops since atchops depends on atlogger now

packages/atchops/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(
   ${CMAKE_CURRENT_LIST_DIR}/src/aes.c
   ${CMAKE_CURRENT_LIST_DIR}/src/aesctr.c
   ${CMAKE_CURRENT_LIST_DIR}/src/base64.c
+  ${CMAKE_CURRENT_LIST_DIR}/src/constants.c
```
xavierchanth, line 11: constants.c contains that runtime lookup map for atchops->mbedtls md types

packages/atchops/include/atchops/constants.h
```diff
@@ -3,4 +3,23 @@
...
+  ATCHOPS_MD_SHA1,
+  ATCHOPS_MD_SHA224,
+  ATCHOPS_MD_SHA256,
+  ATCHOPS_MD_SHA384,
+  ATCHOPS_MD_SHA512,
+  ATCHOPS_MD_SHA3_224,
+  ATCHOPS_MD_SHA3_256,
+  ATCHOPS_MD_SHA3_384,
+  ATCHOPS_MD_SHA3_512,
+} atchops_md_type;
```
xavierchanth, line 21: These are what mbedtls makes available, not sure we need all of them, but I put them anyways. We do have runtime validation for some of them already.

packages/atchops/src/constants.c
```diff
@@ -0,0 +1,10 @@
+#include "atchops/constants.h"
+
+const mbedtls_md_type_t atchops_mbedtls_md_map[] = {
+    [ATCHOPS_MD_NONE] = MBEDTLS_MD_NONE,           [ATCHOPS_MD_MD5] = MBEDTLS_MD_MD5,
+    [ATCHOPS_MD_RIPEMD160] = MBEDTLS_MD_RIPEMD160, [ATCHOPS_MD_SHA1] = MBEDTLS_MD_SHA1,
+    [ATCHOPS_MD_SHA224] = MBEDTLS_MD_SHA224,       [ATCHOPS_MD_SHA256] = MBEDTLS_MD_SHA256,
+    [ATCHOPS_MD_SHA384] = MBEDTLS_MD_SHA384,       [ATCHOPS_MD_SHA512] = MBEDTLS_MD_SHA512,
+    [ATCHOPS_MD_SHA3_224] = MBEDTLS_MD_SHA3_224,   [ATCHOPS_MD_SHA3_256] = MBEDTLS_MD_SHA3_256,
+    [ATCHOPS_MD_SHA3_384] = MBEDTLS_MD_SHA3_384,   [ATCHOPS_MD_SHA3_512] = MBEDTLS_MD_SHA3_512,
+};
```
xavierchanth, line 10: The lookup map, this must be in a `.c` file since it is created at run time. We will get a build error if this is included in a header which is included multiple times, even if that header has #ifndef/#define guards wrapping it.

packages/atchops/include/atchops/constants.h
```diff
@@ -3,4 +3,23 @@
...
+
+extern const mbedtls_md_type_t atchops_mbedtls_md_map[];
```
xavierchanth, line 23: lookup map declaration

packages/atchops/include/atchops/constants.h
JeremyTubongbanua: i think we should only expose the ones we use in the entirety of the atProtocol

i fear someone will come along and think they can just use any algorithm of their choosing (something intern me would do).

xavierchanth: I think it is useful to make these tools available. I believe the only situation where an unsupported hash type should fail right now is in rsa verify, since that has some per-algorithm implementation details.

xavierchanth: We already have guards in place for the hash type:

```
if (mdtype == ATCHOPS_MD_SHA256) {
    hashsize = 32;
  } else {
    // TODO: log error unsupported hash type, untested
    atlogger_log(LOGGER_TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Unsupported hash type for rsa sign\n");
    return 1;
  }
```

I'm going to leave them in for the sake of keeping things in order. Other than rsa sign/verify, the functions in atchops are still able to use these hashing algorithms; they are useful tools to have in the toolbox.

CMakeLists.txt
JeremyTubongbanua: +1

packages/atchops/src/rsa.c
```diff
@@ -13,15 +13,19 @@
...
                      const unsigned char *message, const size_t messagelen, unsigned char *signature) {
   int ret = 1;
 
   size_t hashsize;
-  if (mdtype == MBEDTLS_MD_SHA256) {
+
+  if (mdtype == ATCHOPS_MD_SHA256) {
     hashsize = 32;
   } else {
     // TODO: log error unsupported hash type, untested
```
JeremyTubongbanua, line 27: Remove the // comment, cause the TODO is finished

JeremyTubongbanua (approved):

packages/atchops/include/atchops/constants.h
JeremyTubongbanua: 👍🏼

packages/atchops/src/constants.c
JeremyTubongbanua: +1

PR #297, opened 2024-06-07
feat: monitor hooks, connection send hooks

xavierchanth:
**- What I did**

Implemented the following hooks for monitor connection:

- pre/post notification decryption (perfect for locking the sync client, no race conditions!)

I did a very minimal implementation for monitor, because the other use-cases are not well defined or clear at this time. I went with what I needed and no more than that.

Implemented the following hooks for atclient_connection_send:

- pre/post send
- pre/post receive

I did a full implementation with __mostly__ strong type definitions for the hooks, since I don't foresee any other places we would want them at the connection level. Anywhere else in the connection lifecycle in which the end developer might want intermediate control, they wouldn't need hooks for.

**- How I did it**

**- How to verify it**

feat: monitor hooks, connection send hooks


packages/atclient/include/atclient/connection.h
```diff
@@ -14,9 +14,30 @@
...
+typedef struct atclient_connection_hooks {
+  bool _is_nested_call; // internal variable for preventing infinite recursion (hooks cannot trigger other hooks in
+                        // their nested calls)
+  atclient_connection_send_hook *pre_send;
+  atclient_connection_send_hook *post_send;
+  atclient_connection_send_hook *pre_recv;
+  atclient_connection_send_hook *post_recv;
+  bool readonly_src;
+} atclient_connection_hooks;
+
```
xavierchanth: The hook typedefs for connection

packages/atclient/include/atclient/monitor.h
```diff
@@ -246,6 +246,11 @@ int atclient_monitor_pkam_authenticate(atclient *monitor_conn, const char *atser
  */
 void atclient_monitor_set_read_timeout(atclient *monitor_conn, const int timeout_ms);
 
+typedef struct atclient_monitor_hooks {
+  int (*pre_decrypt_notification)(void);
+  int (*post_decrypt_notification)(void);
+} atclient_monitor_hooks;
+
```
xavierchanth, line 253: Monitor hooks, literally just allow you to run some code in between steps in the monitor read function, unlike the connection, they don't expose any internal values.

packages/atclient/src/connection.c
```diff
@@ -204,29 +202,77 @@ exit: {
...
+    if (src == NULL) {
+      atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to allocate memory for src\n");
+      allocate_src = false; // don't try to free since the memory failed to be allocated
+      goto exit;
+    }
+    memcpy(src, src_r, srclen);
+  } else {
+    src = (unsigned char *)src_r;
+  }
+
```
xavierchanth, line 229: src is conditionally copied & allocated as an editable buffer, but only if readonly_src = false. This allows us to maintain performance by default, but allows the end developer to take control if they really want to.

packages/atclient/include/atclient/connection.h
```diff
@@ -42,6 +65,9 @@ typedef struct atclient_connection {
  * @param type the type of connection to initialize,
  * if it is ATCLIENT_CONNECTION_TYPE_ROOT, then '\\n' will be used to check if it is connected.
  * if it is ATCLIENT_CONNECTION_TYPE_ATSERVER, then 'noop:0\r\n' will be used to check if it is connected
+ *
+ * @note hooks have a separate init method, called atlcient_connection_hooks_init. After this method, call
```
JeremyTubongbanua: typo `atlcient_connection_hooks_init`

packages/atclient/include/atclient/connection.h
```diff
@@ -105,4 +131,31 @@ void atclient_connection_free(atclient_connection *ctx);
  */
 int atclient_connection_get_host_and_port(atclient_atstr *host, int *port, const atclient_atstr url);
 
+/**
+ * @brief Initialize the hooks memory allocation
+ *
+ * @param ctx the struct for the connection
+ */
+void atclient_connection_hooks_init(atclient_connection *ctx);
+
```
JeremyTubongbanua: We have `atclient_connection_hooks_init` function, but no `atclient_connection_hooks_free` function

JeremyTubongbanua (commented):
I would like to propose this flow for initializing hooks and setting them to a client.

```c
/* 1. Initialize clients */
atclient monitor_client;
atclient_init(&monitor_client);

atclient main_client;
atclient_init(&main_client);

/* 2. Initialize connection hooks */
atclient_connection_hooks my_hooks;
atclient_connection_hooks_init(&my_hooks);

/* 3. Set hooks to clients */
atclient_connection_set_hooks(&(monitor_client->atserver_connection), &my_hooks);
atclient_connection_set_hooks(&(main_client->atserver_connection), &my_hooks);

/* 4. Do stuff ... */

/* 5. Free everything */
atclient_free(&monitor_client);
atclient_connection_hooks_free(&my_hooks); // it doesn't matter when hooks is freed in relation to its connections, since it is allocated independently from the connection contexts themselves
atclient_free(&main_client); 
```

packages/atclient/include/atclient/monitor.h
```diff
@@ -246,6 +246,11 @@ int atclient_monitor_pkam_authenticate(atclient *monitor_conn, const char *atser
  */
 void atclient_monitor_set_read_timeout(atclient *monitor_conn, const int timeout_ms);
 
+typedef struct atclient_monitor_hooks {
+  int (*pre_decrypt_notification)(void);
```
JeremyTubongbanua, line 250: could we typedef this like:

```c
typedef int(atclient_monitor_decrypt_notification_hook)(void);
```

So the struct would look like

```h
typedef struct atclient_monitor_hooks {
    atclient_monitor_decrypt_notification_hook *pre;
    atclient_monitor_decrypt_notification_hook *post;
}
```

JeremyTubongbanua:
Upon approval, we should write up a ticket to write tests for this feature

packages/atclient/include/atclient/connection.h
xavierchanth: We don't need one, the hooks will be freed by atclient_connection_free

packages/atclient/include/atclient/monitor.h
xavierchanth: I originally had done that, but: If the type isn't complicated, I want to avoid typedefs, since if we decide to add parameters to the signature in the future, then we need to make more breaking changes than otherwise necessary.

xavierchanth:
> I would like to propose this flow for initializing hooks and setting them to a client.
> 
> ```c
> /* 1. Initialize clients */
> atclient monitor_client;
> atclient_init(&monitor_client);
> 
> atclient main_client;
> atclient_init(&main_client);
> 
> /* 2. Initialize connection hooks */
> atclient_connection_hooks my_hooks;
> atclient_connection_hooks_init(&my_hooks);
> 
> /* 3. Set hooks to clients */
> atclient_connection_set_hooks(&(monitor_client->atserver_connection), &my_hooks);
> atclient_connection_set_hooks(&(main_client->atserver_connection), &my_hooks);
> 
> /* 4. Do stuff ... */
> 
> /* 5. Free everything */
> atclient_free(&monitor_client);
> atclient_connection_hooks_free(&my_hooks); // it doesn't matter when hooks is freed in relation to its connections, since it is allocated independently from the connection contexts themselves
> atclient_free(&main_client); 
> ```

Hooks has to be part of the memory of the atclient_connection struct, otherwise we have to pass around a new object.

The reason there is only an init and not a free, is that we don't want hooks to be using more memory than otherwise necessary. So the hooks init is a second init on the atclient_connection struct which additionally allocates memory for the hooks struct. This should still all be freed by atclient_connection, so that you never have to worry about previous state of an atclient_connection in order to free it.

Should I call it enable hooks instead?

packages/atclient/src/connection.c
```diff
@@ -344,6 +432,10 @@ void atclient_connection_free(atclient_connection *ctx) {
   memset(ctx->host, 0, ATCLIENT_CONSTANTS_HOST_BUFFER_SIZE);
   ctx->port = -1;
   ctx->should_be_connected = false;
+
+  if (ctx->hooks != NULL) {
+    free(ctx->hooks);
+  }
```
xavierchanth, line 438: FYI here is where the hooks are freed, this is allocated by `atclient_connection_enable_hooks`

xavierchanth: you don't need to worry about the memory of `ctx->hooks->XYZ` since they are function pointers.

packages/atclient/include/atclient/monitor.h
JeremyTubongbanua: 👍🏼 Ok

The reason I brought it up is cause it was a little confusing for me to understand at first. It was hard for me to realize that it wasn't anything different than what we had going on in `connection_send`

But I also feel like it is okay to keep the breaking change in for now, in order to stay uniform with what we have in `connection_send` as well.

JeremyTubongbanua:
> > I would like to propose this flow for initializing hooks and setting them to a client.
> > ```c
> > /* 1. Initialize clients */
> > atclient monitor_client;
> > atclient_init(&monitor_client);
> > 
> > atclient main_client;
> > atclient_init(&main_client);
> > 
> > /* 2. Initialize connection hooks */
> > atclient_connection_hooks my_hooks;
> > atclient_connection_hooks_init(&my_hooks);
> > 
> > /* 3. Set hooks to clients */
> > atclient_connection_set_hooks(&(monitor_client->atserver_connection), &my_hooks);
> > atclient_connection_set_hooks(&(main_client->atserver_connection), &my_hooks);
> > 
> > /* 4. Do stuff ... */
> > 
> > /* 5. Free everything */
> > atclient_free(&monitor_client);
> > atclient_connection_hooks_free(&my_hooks); // it doesn't matter when hooks is freed in relation to its connections, since it is allocated independently from the connection contexts themselves
> > atclient_free(&main_client); 
> > ```
> 
> Hooks has to be part of the memory of the atclient_connection struct, otherwise we have to pass around a new object.
> 
> The reason there is only an init and not a free, is that we don't want hooks to be using more memory than otherwise necessary. So the hooks init is a second init on the atclient_connection struct which additionally allocates memory for the hooks struct. This should still all be freed by atclient_connection, so that you never have to worry about previous state of an atclient_connection in order to free it.
> 
> Should I call it enable hooks instead?

Yes, enable hooks makes more sense to the flow we have going on in the C SDK.

If we call it `init`, then it tells the caller that they have to at some point call `free`. But calling it `enable_hooks` solves that problem.

packages/atclient/include/atclient/connection.h
```diff
@@ -14,9 +14,30 @@
...
 
+typedef int(atclient_connection_send_hook)(const unsigned char *src, const size_t srclen, unsigned char *recv,
+                                           const size_t recvsize, size_t *recvlen);
+
+typedef enum atclient_connection_hook_type {
+  ACHT_NONE = 0,
+  ACHT_PRE_SEND,
+  ACHT_POST_SEND,
+  ACHT_PRE_RECV,
+  ACHT_POST_RECV,
```
JeremyTubongbanua: Would it be possible to expand the acronym to `ATCLIENT_CONNECTION_HOOK_TYPE_*`?

Will this enum be typed by the caller enough times to justify having this be shortened?

I think we should expand it for 3 reasons:
1. intellisense
2. less confusing on what `ACHT` means
3. guaranteed namespacing (there's a chance someone will take `ACHT_*` but no chance someone will take `ATCLIENT_CONNECTION_HOOK_TYPE*`

JeremyTubongbanua (approved):

PR #300, opened 2024-06-10
fix: network related hooks fixes

xavierchanth:
**- What I did**

- connection_send hooks would cause a segfault when the network is down since the hooks would get wiped
  - put guards for that in place.
- similarly, I also had to make sure that hooks are preserved across network connection resets

- the monitor post_decrypt hooks is used to unlock the worker atclient, but it doesn't run if the decrypt notification fails:
  - fixed such that the hook always runs, and the hook provides the return value of the decrypt notification call

**- How I did it**

**- How to verify it**

fix: network related hooks fixes

### DIFF
--- a/packages/atclient/include/atclient/monitor.h
+++ b/packages/atclient/include/atclient/monitor.h
@@ -248,7 +248,7 @@ void atclient_monitor_set_read_timeout(atclient *monitor_conn, const int timeout
 
 typedef struct atclient_monitor_hooks {
   int (*pre_decrypt_notification)(void);
-  int (*post_decrypt_notification)(void);
+  int (*post_decrypt_notification)(int);
 } atclient_monitor_hooks;
 
 /**
--- a/packages/atclient/src/atclient.c
+++ b/packages/atclient/src/atclient.c
@@ -1,13 +1,13 @@
 #include "atclient/atclient.h"
-#include "atclient/constants.h"
-#include "atclient/atclient_utils.h"
 #include "atchops/base64.h"
 #include "atchops/rsa.h"
 #include "atclient/atclient.h"
+#include "atclient/atclient_utils.h"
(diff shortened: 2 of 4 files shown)


packages/atclient/src/atclient.c
```diff
@@ -227,13 +227,22 @@ void atclient_set_read_timeout(atclient *ctx, int timeout_ms) {
...
   ctx->atserver_connection_started = false;
   memset(&(ctx->atserver_connection), 0, sizeof(atclient_connection));
 
+  // (re) initialize the atserver connection
   atclient_connection_init(&(ctx->atserver_connection), ATCLIENT_CONNECTION_TYPE_ATSERVER);
   ctx->atserver_connection_started = true;
 
+  // add back hooks
+  ctx->atserver_connection.hooks = conn_hooks;
+
```
xavierchanth: Originally tried implementing a reset function to replace with this, but it caused a slew of other problems. 

I found the simplest solution is to save the address to the hooks struct and remove it from the ctx, that way it isn't overwritten by _free or _init. Then we simply write the address to the hooks back, this is also naturally resilient to the case where hooks are NULL.

packages/atclient/include/atclient/monitor.h
```diff
@@ -248,7 +248,7 @@ void atclient_monitor_set_read_timeout(atclient *monitor_conn, const int timeout
 
 typedef struct atclient_monitor_hooks {
   int (*pre_decrypt_notification)(void);
-  int (*post_decrypt_notification)(void);
+  int (*post_decrypt_notification)(int);
```
xavierchanth, line 251: changed to int which provides the return value of the decryption_notification step

JeremyTubongbanua (dismissed):

JeremyTubongbanua (approved):

PR #318, opened 2024-06-14
fix: atclient_notify memory leaks

JeremyTubongbanua:
**- What I did**
- Refactored metadata internal functionality
- There would be 25 definitely lost bytes that would be lost every time a atclient_notify_params would be allocated

**- How to verify it**
at_talk valgrind memory leak summary after sending a notification (at talk message)
```
==83172== LEAK SUMMARY:
==83172==    definitely lost: 0 bytes in 0 blocks
==83172==    indirectly lost: 0 bytes in 0 blocks
==83172==      possibly lost: 272 bytes in 1 blocks
==83172==    still reachable: 269,960 bytes in 251 blocks
==83172==         suppressed: 0 bytes in 0 blocks
```

fix: atclient_notify memory leaks

### DIFF
--- a/examples/desktop/at_talk/src/main.c
+++ b/examples/desktop/at_talk/src/main.c
@@ -165,6 +165,9 @@ int main(int argc, char *argv[]) {
     pthread_mutex_unlock(&client_mutex);
 
     printf("%s%s%s: ", HBLU, from_atsign, reset);
+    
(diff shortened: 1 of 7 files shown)


xavierchanth (approved):
Thanks for working on this change

PR #321, opened 2024-06-14
feat: rsakey deep clone

xavierchanth:
**- What I did**

- added `atchops_rsakey_publickey_clone` & `atchops_rsakey_privatekey_clone` functions

FYI: I chose to make it create memory rather than depending on init, so that it is resilient if we support multiple key sizes in the future

**- How I did it**

**- How to verify it**

feat: rsakey deep clone

### DIFF
--- a/packages/atchops/include/atchops/rsakey.h
+++ b/packages/atchops/include/atchops/rsakey.h
@@ -4,7 +4,7 @@
 #include <stddef.h>
 
 typedef struct atchops_rsakey_param {
-  size_t len;    // length of the number in bytes
+  size_t len;           // length of the number in bytes
   unsigned char *value; // hex byte array of the number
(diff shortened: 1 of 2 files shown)


JeremyTubongbanua (approved):

JeremyTubongbanua:
Why is it called a deep clone?

And when would this ever be useful?

JeremyTubongbanua:
Oops sorry, accidentally merged..

xavierchanth:
> Why is it called a deep clone?
> 
> And when would this ever be useful?

I called the method clone, but it is deep clone of the memory as opposed to a shallow clone of just addresses. I'm using it in the c daemon to create a copy of the signing key, since for some reason, I was getting a race condition if I use the atkeys in the atserver directly (even though things are only reading from it)

PR #322, opened 2024-06-14
fix: atlogger memory leak

xavierchanth:
**- What I did**
- Added a function for freeing memory created by the atlogger singleton
- It isn't mandatory that you use this method, since the amount allocated will never change and 99% of the time you will want to free this memory, it is right as the program exits. But for the sake of having a way to free every single piece of memory properly, we need this function.

**- How I did it**

**- How to verify it**

fix: atlogger memory leak


JeremyTubongbanua:
Do you think statically allocating the singleton would be better? It is unintuitive to me to have to call `atlogger_free` at the end of every single application I would ever write using the C SDK

xavierchanth:
I tried that first couldn't think of a good way to do it, after thinking about it from afar, I have a good way to do it. I'll mark this PR as draft and open a fresh history without this stuff added. Once I have the new PR I will discard this branch.

PR #324, opened 2024-06-17
chore: use static memory for at_logger

xavierchanth:
closes #322 

**- What I did**

**- How I did it**

**- How to verify it**

chore: use static memory for at_logger


xavierchanth:
Note: Not necessarily optimal, but it gets the job done, doesn't break the public API, and leaves room for us to revert back to dynamic allocation if we need to

JeremyTubongbanua (approved):
Did my own manual test of this branch and functions

PR #328, opened 2024-06-29
fix: atclient_create_shared_encryption_key_pair_for_me_and_other memory leaks

JeremyTubongbanua:
**- What I did**
- atchops_rsa_publickey was not freed, so it is now freed
- new `/get` repl feature

**- How to verify it**

### DIFF
--- a/examples/desktop/repl/src/main.c
+++ b/examples/desktop/repl/src/main.c
@@ -102,19 +102,82 @@ int main(int argc, char *argv[]) {
 
   bool loop = true;
   do {
-    printf("Enter command: \n");
+    printf("Enter command: ");
     fgets(buffer, buffersize, stdin);
     bufferlen = strlen(buffer);
-    if (strncmp(buffer, "/exit", strlen("/exit")) == 0) {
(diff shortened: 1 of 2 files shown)


realvarx (dismissed):
LGTM

Just a small thing (not from this PR, but maybe you can add it to it), this isnt self enc key, right? Its just a tweak to the comment:
https://github.com/atsign-foundation/at_c/blob/6b893d51493424c92cc4c9286613bbeba0e4d9ae/packages/atclient/src/encryption_key_helpers.c#L285-L288

JeremyTubongbanua:
> LGTM
> 
> Just a small thing (not from this PR, but maybe you can add it to it), this isnt self enc key, right? Its just a tweak to the comment:
> 
> https://github.com/atsign-foundation/at_c/blob/6b893d51493424c92cc4c9286613bbeba0e4d9ae/packages/atclient/src/encryption_key_helpers.c#L285-L288

Thank you, will make this change and ask for re-review

realvarx (approved):

PR #355, opened 2024-07-30
chore: 3.24 minimum cmake version

### LINKED ISSUE #338: Update CMake minimum required version to 3.24
Features that use 3.24:
- Policy CMP0135
- find_package's fetch content redirection search mode

Debian derivatives are problematic with this version of cmake since the system packages are older versions. The solution is to use pip to install cmake, but you may also need to install to a venv if you are on debian 12, as system packages are not supported with pip.

JeremyTubongbanua:
closes #338 

chore: 3.24 minimum cmake version


cpswan (approved):

xavierchanth (approved):

xavierchanth:
This change is also in #357 btw

PR #367, opened 2024-08-09
feat: copy everything in release instead of `packages/*`

JeremyTubongbanua:
**- What I did**

**- How I did it**

**- How to verify it**

feat: copy everything in release instead of `packages/*`


cpswan:
@JeremyTubongbanua there are some expectations around source tarballs (for things like OpenWRT makefiles) that drove things towards using the contents of packages.

Also by default any release on GitHub will have a snapshot of the repo source at that point in time.

PR #381, opened 2024-08-22
fix: Add atclient_pkam_authenticate_options and populate atserver host and…

### LINKED ISSUE #285: at_c: passing a `NULL` into `atserver_host` and `0` into `atserver_port` will use `atclient_utils_find_atserver_address`
This is a nice QOL feature.

- passing a NULL into atserver_host
- passing 0 into atserver_port 
- will use atclient_utils_find_atserver_address

So that the person using `atclient_pkam_authenticate` won't have to manage atserver_host and atserver_port

sitaram-kalluri:
… port when values are not set

**- What I did**
- Modify `atclient_pkam_authenticate` function to fetch the atserver host and port when not set by the user. 

**- How I did it**
- Introduce 'atclient_pkam_authenticate_options' structure which contains root server host and port and atserver host and port as data members.
  -  The root server host and port will be defaulted to the prod root server.
- Modify the signature of `atclient_pkam_authenticate` function to accept 'atclient_pkam_authenticate_options' and remove atserver_host and atserver_port.
- Refactor the examples to consume the `atclient_apkam_authenticate` function changes.

**- How to verify it**
- Tested manually and able to fetch the atserver host and port when not set by the user.
- After the changes, ran examples and works expected.


packages/atclient/include/atclient/atclient.h
```diff
@@ -123,8 +123,7 @@ void atclient_stop_atserver_connection(atclient *ctx);
  * @param atsign the atsign the atkeys belong to, this string is assumed to be null terminated
  * @return int 0 on success, non-zero on error
  */
-int atclient_pkam_authenticate(atclient *ctx, const char *atserver_host, const int atserver_port,
-                               const atclient_atkeys *atkeys, const char *atsign);
+int atclient_pkam_authenticate(atclient *ctx, const atclient_atkeys *atkeys, const char *atsign, atclient_pkam_authenticate_options *options);
```
JeremyTubongbanua: I would make the function signature this:

```
int atclient_pkam_authenticate(atclient *ctx, const char *atsign, const atclient_atkeys *atkeys,, atclient_pkam_authenticate_options *options);
```

Because `atkeys` may become an optional argument in the future, so let's put it near the end.

packages/atclient/src/request_options.c
```diff
@@ -8,6 +8,10 @@
 
 #define TAG "request_options"
 
+// default param values for ATCLIENT PKAM AUTHENTICATE OPTIONS
+#define ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_ROOT_SERVER_HOST "root.atsign.org"
+#define ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_ROOT_SERVER_PORT 64
```
JeremyTubongbanua, line 14: These should be in `constants.h`

packages/atclient/src/request_options.c
```diff
@@ -1426,3 +1433,319 @@ void atclient_get_atkeys_request_options_unset_show_hidden(atclient_get_atkeys_r
   options->show_hidden = false;
   atclient_get_atkeys_request_options_set_show_hidden_initialized(options, false);
 }
+
+void atclient_pkam_authenticate_options_init(atclient_pkam_authenticate_options *options) {
+  memset(options, 0, sizeof(atclient_pkam_authenticate_options));
+  options->root_server_host = ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_ROOT_SERVER_HOST;
+  options->root_server_port = ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_ROOT_SERVER_PORT;
```
JeremyTubongbanua: Also, we don't call it the "root" server anymore. We call it the "atDirectory", and ideally, our new SDK is aligned with our new marketing terms

packages/atclient/include/atclient/request_options.h
```diff
@@ -109,6 +118,18 @@ typedef struct atclient_get_atkeys_request_options {
...
+/*
+ * 5. Pkam auhtenticate Request Options
+ */
+
+typedef struct atclient_pkam_authenticate_options {
+  char *root_server_host;
+  int root_server_port;
+  char *atserver_host;
+  int atserver_port;
+  uint8_t _initialized_fields[1];
```
JeremyTubongbanua: ports have a range of 0-65535

So instead of using `int`, we could use `uint16_t`

JeremyTubongbanua (commented):
I had some suggestions. They are small nit picks, so it is not a huge priority.

packages/atclient/src/atclient.c
```diff
@@ -186,19 +186,30 @@ bool atclient_is_atsign_initialized(const atclient *ctx) {
...
+  if ((ret = atclient_pkam_authenticate_validate_arguments(ctx, atkeys, atsign)) != 0) {
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "atclient_pkam_authenticate_validate_arguments: %d\n", ret);
     return ret;
   }
 
+  /*
+  * 1.1 If atserver_host and atserver_port are null and 0 respectively, then fetch the server host and port.
+  */
+  if(options->atserver_host == NULL && options->atserver_port == 0){
+    atclient_utils_find_atserver_address(options->root_server_host, options->root_server_port, atsign, &(options->atserver_host), &(options->atserver_port));
```
JeremyTubongbanua: This function `atclient_utils_find_atserver_address` returns an error code which should be handled.

packages/atclient/src/atclient.c
```diff
@@ -186,19 +186,30 @@ bool atclient_is_atsign_initialized(const atclient *ctx) {
...
+  }
+  printf("Server Host and port : %s,%d", options->atserver_host, options->atserver_port);
```
JeremyTubongbanua: Use `atlogger` instead of `printf`

packages/atclient/src/atclient.c
```diff
@@ -244,7 +255,7 @@ int atclient_pkam_authenticate(atclient *ctx, const char *atserver_host, const i
   /*
    * 4. Start atServer connection (kill the existing connection if it exists)
    */
-  if ((ret = atclient_start_atserver_connection(ctx, atserver_host, atserver_port)) != 0) {
+  if ((ret = atclient_start_atserver_connection(ctx, options->atserver_host, options->atserver_port)) != 0) {
```
JeremyTubongbanua, line 268: Before calling this function, we should check if `options->atserver_host` and `options->atserver_port` were set by the caller.

```c
if(!atclient_pkam_authenticate_options_is_atserver_host_initialized(&options) || !atclient_pkam_authenticate_options_Is_atserver_port_initialized(&options)) {
   // handle error
}
```

JeremyTubongbanua (commented):
Some more changes which I think are more necessary

packages/atclient/include/atclient/atclient.h
sitaram-kalluri: Moved the atKeys after the atSign.

packages/atclient/src/request_options.c
sitaram-kalluri: Moved to constants.h

packages/atclient/src/request_options.c
sitaram-kalluri: Modified the variable as "ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_AT_DIRECTORY_HOST" and "ATCLIENT_PKAM_AUTHENTICATE_DEFAULT_AT_DIRECTORY_PORT"

packages/atclient/include/atclient/request_options.h
sitaram-kalluri: Update the type to uint16_t

packages/atclient/src/atclient.c
sitaram-kalluri: If the return code is not 0, then return.

packages/atclient/src/atclient.c
sitaram-kalluri: Removed the printf statement

packages/atclient/src/atclient.c
sitaram-kalluri: If the atserver_host and atserver_port are not initialized, then goto exit.

examples/desktop/at_talk/src/main.c
```diff
@@ -93,22 +90,15 @@ int main(int argc, char *argv[]) {
...
   /*
    * 4. Authenticate client connection (for crud operations)
    */
   pthread_mutex_lock(&client_mutex);
-  if ((ret = atclient_pkam_authenticate(&atclient1, atserver_host, atserver_port, &atkeys, from_atsign)) != 0) {
+  atclient_pkam_authenticate_options options;
+  atclient_pkam_authenticate_options_init(&options);
+  if ((ret = atclient_pkam_authenticate(&atclient1, from_atsign, &atkeys, &options)) != 0) {
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "\natclient_pkam_authenticate: %d\n", ret);
     goto exit;
```
JeremyTubongbanua, line 103: Regarding all the examples/

Is passing `NULL` possible into the `options`?

packages/atclient/include/atclient/request_options.h
```diff
@@ -36,6 +36,15 @@
 #define ATCLIENT_GET_ATKEYS_REQUEST_OPTIONS_REGEX_INITIALIZED (VALUE_INITIALIZED << 0)
 #define ATCLIENT_GET_ATKEYS_REQUEST_OPTIONS_SHOW_HIDDEN_INITIALIZED (VALUE_INITIALIZED << 1)
 
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_AT_DIRECTORY_HOST_INDEX 0
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_AT_DIRECTORY_PORT_INDEX 0
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_ATSERVER_HOST_INDEX 0
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_ATSERVER_PORT_INDEX 0
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_AT_DIRECTORY_HOST_INITIALIZED (VALUE_INITIALIZED << 0)
+#define ATCLIENT_PKAM_AUTHENTICATE_OPTIONS_AT_DIRECTORY_PORT_INITIALIZED (VALUE_INITIALIZED << 1)
```
JeremyTubongbanua, line 44: `AT_DIRECTORY` should be `ATDIRECTORY`

packages/atclient/src/atclient.c
```diff
@@ -178,27 +179,41 @@ bool atclient_is_atsign_initialized(const atclient *ctx) {
...
   }
 
+  /*
+   * 1.1 If atserver_host and atserver_port are null and 0 respectively, then fetch the server host and port.
+   */
+  if (options->atserver_host == NULL && options->atserver_port == 0) {
+    if (atclient_utils_find_atserver_address(options->at_directory_host, options->at_directory_port, atsign,
+                                             &(options->atserver_host), &(options->atserver_port)) != 0) {
+      return ret;
+    }
```
JeremyTubongbanua, line 214: If an error happens here, then `atclient_pkam_authenticate_options_free` will never be called.

(This isn't written down in our Contributing spec yet, my apologies), but we have an ongoing convention where for every `*_init` function that is called, `*_free` must be called at the end of the variable's lifetime. This is for future proofing and easy memory management.

packages/atclient/src/request_options.c
JeremyTubongbanua: You might've forgot to delete it from `request_options.c`

JeremyTubongbanua (approved):
I have more nits

But this works! :)

PR #385, opened 2024-08-27
fix: implment new method of trimming 'data:' from server responses

### LINKED ISSUE #288: at_c: Use `data:` token instead of checking first 5 characters
image

@XavierChanth Found bug where the bytes read had some leading excess bytes.

Within the `atclient_connection_send` function: Instead of checking if the first 5 bytes is `data:`, we should instead look for the token `data:` then truncate anything before that, also including the prefix itself.

For example:

If my received bytes was `@colin@data:abcdefgh`, then I would only be concerned about `abcdefgh` and `@colin@data:` should be removed from the string.

`encryption_key_helpers` and `atclient_monitor_read` should also be checked because they may use some raw `mbedtls_ssl_read` functions there.

Anywhere that does something like: `atclient_string_utils_starts_with(recv, "data:")` should be modified with the new solution

srieteja:
Closes https://github.com/atsign-foundation/at_c/issues/288

**- What I did**
- fix the issue: 'bytes read had some leading excess bytes'

**- How I did it**
- Introduce a method in string_utils.c named "atclient_string_utils_get_substring_position()" that locates the position of a given substring.
- Using the above method to locate the 'data:' substring and then moving the pointer 5 positions ahead to skip the 'data:' substring.
- Replaced existing method of trimming 'data:' prefix if a server response starts with 'data:', with the new implementation

**- How to verify it**
- Successful build should indicate no compile-time errors. Yet to do a manual test to verify individual functionalities updated with changes in this PR.

fix: implment new method of trimming 'data:' from server responses


JeremyTubongbanua (dismissed):

srieteja:
I will merge this once I can test the changes

packages/atclient/src/atclient.c
```diff
@@ -432,6 +432,7 @@ int atclient_send_heartbeat(atclient *heartbeat_conn) {
   /*
    * 3. Parse response
    */
+  // how about just doing ptr == "data:ok" ?
```
srieteja, line 435: Can the following snipped replaced with `ptr == "data:ok"` or is there a reason for doing it this way?

packages/atclient/include/atclient/string_utils.h
```diff
@@ -37,6 +37,16 @@ bool atclient_string_utils_starts_with(const char *string, const char *prefix);
  */
 bool atclient_string_utils_ends_with(const char *string, const char *suffix);
 
+/**
+ *@brief locates the position of a subtring in a given string
```
JeremyTubongbanua: - add space before `@brief`
- `subtring` -> `substring`

packages/atclient/src/atclient.c
JeremyTubongbanua: `ptr` is an address, not a value.

The value of `ptr` is a byte address that could look something like this: `0xb4aca`, so it would always return false.

You would have to use `strcmp` which compares every byte up to the null-terminator.

packages/atclient/src/atclient.c
```diff
@@ -432,6 +432,7 @@ int atclient_send_heartbeat(atclient *heartbeat_conn) {
...
   if (!atclient_string_utils_starts_with((const char *)ptr, "data:ok") &&
       !atclient_string_utils_ends_with((const char *)ptr, "data:ok")) {
```
JeremyTubongbanua: I feel like this could be changed to `strcmp`

but feel free to do what you like

packages/atclient/src/atclient_get_public_key.c
```diff
@@ -81,16 +81,17 @@ int atclient_get_public_key(atclient *atclient, atclient_atkey *atkey, char **va
...
-  if (!atclient_string_utils_starts_with(response, "data:")) {
+  char *response_trimmed = NULL;
+  // below method points the response_trimmed variable to the position of 'data:' substring
+  if(atclient_string_utils_get_substring_position(response, DATA_TOKEN, &response_trimmed) != 0) {
     ret = 1;
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "recv was \"%.*s\" and did not have prefix \"data:\"\n",
                  (int)recv_len, recv);
     goto exit;
   }
+  response_trimmed = response_trimmed + 5; // +5 to skip the "data:" prefix
```
JeremyTubongbanua: something we could do here is

`response_trimmed = response_trimmed + strlen(DATA_TOKEN)` for future-proofing sake.

This would be optimized by the compiler to be changed to `5`, so it is not unoptimal, but this solution is chill too

packages/atlogger/src/atlogger.c
```diff
@@ -10,7 +10,7 @@
 #define PREFIX_BUFFER_LEN 64
 #define INFO_PREFIX "\e[0;32m[INFO]\e[0m"
 #define WARN_PREFIX "\e[0;31m[WARN]\e[0m"
-#define ERROR_PREFIX "\e[1;31m[ERRO]\e[0m"
+#define ERROR_PREFIX "\e[1;31m[ERROR]\e[0m"
```
JeremyTubongbanua, line 13: I think making `ERRO` longer will no longer align the logs nicely

All of the other tags are the same length of 4 characters long,

But if you've tested it and it looks nice then we can keep it

JeremyTubongbanua (dismissed):

packages/atlogger/src/atlogger.c
srieteja: 'ERRO' would align, but users would think that's a typo. We could use something like 'ERRR', what do you think?

packages/atclient/src/atclient_get_public_key.c
srieteja: Your solution looks better as it would make the code more readable, let me update that in the code

srieteja: This change has been made in [2795073](https://github.com/atsign-foundation/at_c/pull/385/commits/279507393dc0f61b49b9e177c30cef390bc3145b)

packages/atclient/src/atclient.c
srieteja: This change has been made in [2795073](https://github.com/atsign-foundation/at_c/pull/385/commits/279507393dc0f61b49b9e177c30cef390bc3145b)

packages/atclient/include/atclient/string_utils.h
srieteja: This change has been made in [2795073](https://github.com/atsign-foundation/at_c/pull/385/commits/279507393dc0f61b49b9e177c30cef390bc3145b)

packages/atclient/src/string_utils.c
```diff
@@ -59,6 +59,25 @@ bool atclient_string_utils_ends_with(const char *string, const char *suffix) {
...
+  if(strlen(substring) > strlen(string)) {
+    ret = -1;
+    goto exit;
+  }
+  if(position == NULL) {
+    ret = -1;
+    goto exit;
+  }
+  *position =  strstr(string, substring);
+  if(*position == NULL) {
```
xavierchanth, line 73: Can you verify this call please?

position should be a pointer to an integer meaning that `*position == NULL` will be true if the integer value is zero (NULL = 0 on most (all?) platforms)

I think you mean `position == NULL` here

packages/atlogger/src/atlogger.c
xavierchanth: Could add a space after the other ones

xavierchanth (commented):
Other than my one concern looks good

packages/atlogger/src/atlogger.c
JeremyTubongbanua: Anything works! Not a huge deal

packages/atclient/src/string_utils.c
srieteja: We do not deal with any integers here. The method _strstr()_ returns a pointer to the location of the substring. So in this case, _position_ is a pointer to the character 'd'(as in _data:_) wherever it is located in the given string. So *position would actually resolve to the string starting with _data:_ until terminated. If you find the variable name _position_ misleading LMK, so that I can find an alternative that is better suited.

packages/atclient/src/atclient.c
srieteja: The change has been made to replace existing if-statement with strcmp in [2795073](https://github.com/atsign-foundation/at_c/pull/385/commits/279507393dc0f61b49b9e177c30cef390bc3145b)

packages/atclient/src/string_utils.c
srieteja: position is a double pointer, so *position is actually a pointer (points to 'd' in data:). **position would resolve to the letter 'd'. This would only return true if *position does not point to anything.

packages/atlogger/src/atlogger.c
srieteja: In that case, I'm leaving it as 'ERROR' so that it's clearer.

xavierchanth (approved):

PR #402, opened 2024-09-19
chore: reuse existing functions for calculating protocol fragment

xavierchanth:
**- What I did**

We have methods to calculate the string length of individual metadata fragments, as well as the entire metadata fragment. (fragment as in fragment of an at_protocol verb)

I replaced implementation in the function for calculating entire string length with calls to the individual functions instead of doing each calculation manually.

**- How I did it**

**- How to verify it**

See https://github.com/atsign-foundation/noports/pull/1363

chore: reuse existing functions for calculating protocol fragment

### DIFF
--- a/packages/atclient/src/metadata.c
+++ b/packages/atclient/src/metadata.c
@@ -1310,8 +1310,6 @@ int atclient_atkey_metadata_to_protocol_str(const atclient_atkey_metadata *metad
    * 3. Do a sanity check
    */
 
-  printf("pos: %lu\nstrlen(*metadata_str): %lu\nexpected_metadatastr_len: %lu\n", pos, strlen(*metadata_str),
-         expected_metadatastr_len);
   if (strlen(*metadata_str) != expected_metadatastr_len) {
     ret = 1;
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR,
--- a/packages/atclient/src/metadata.c
+++ b/packages/atclient/src/metadata.c
@@ -1004,10 +1004,10 @@ size_t atclient_atkey_metadata_ttl_strlen(const atclient_atkey_metadata *metadat
 }
 
 size_t atclient_atkey_metadata_ttb_strlen(const atclient_atkey_metadata *metadata) {
-  if(metadata == NULL) {
+  if (metadata == NULL) {
     return 0;
   }
-  if(!atclient_atkey_metadata_is_ttb_initialized(metadata)) {
+  if (!atclient_atkey_metadata_is_ttb_initialized(metadata)) {
     return 0;
   }
   return strlen(":ttb:") // :ttb:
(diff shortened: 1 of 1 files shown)


packages/atclient/src/metadata.c
```diff
@@ -1199,8 +1199,9 @@ int atclient_atkey_metadata_to_protocol_str(const atclient_atkey_metadata *metad
   /*
    * 2. Create protocol string
    */
-  const size_t metadata_str_size = atclient_atkey_metadata_protocol_strlen(metadata) + 1;
-  const size_t expected_metadatastr_len = metadata_str_size - 1;
+  const size_t expected_metadatastr_len = atclient_atkey_metadata_protocol_strlen(metadata);
+  const size_t metadata_str_size = expected_metadatastr_len + 1;
```
xavierchanth, line 1203: flipped these variables so we don't have to +1 then -1

packages/atclient/src/metadata.c
```diff
@@ -1210,112 +1211,114 @@ int atclient_atkey_metadata_to_protocol_str(const atclient_atkey_metadata *metad
   memset(*metadata_str, 0, sizeof(char) * metadata_str_size);
 
   if (atclient_atkey_metadata_is_ttl_initialized(metadata)) {
-    sprintf(*metadata_str + pos, ":ttl:%ld", metadata->ttl);
-    pos += 5 + atclient_string_utils_long_strlen(metadata->ttl);
+    sprintf((*metadata_str) + pos, ":ttl:%ld", metadata->ttl);
+    pos += atclient_atkey_metadata_ttl_strlen(metadata);
```
xavierchanth, line 1215: pattern 1: drop in replacement of manual calculation for the method

packages/atclient/src/metadata.c
```diff
@@ -1210,112 +1211,114 @@ int atclient_atkey_metadata_to_protocol_str(const atclient_atkey_metadata *metad
...
     if (metadata->ccd) {
-      sprintf(*metadata_str + pos, ":ccd:true");
-      pos += 9;
+      sprintf((*metadata_str) + pos, ":ccd:true");
     } else {
-      sprintf(*metadata_str + pos, ":ccd:false");
-      pos += 10;
+      sprintf((*metadata_str) + pos, ":ccd:false");
     }
+    pos += atclient_atkey_metadata_ccd_strlen(metadata);
```
xavierchanth, line 1234: pattern 2: we internally call this same if statement inside `atclient_atkey_metadata_<thing>_strlen` for all boolean metadata strlen functions, so we can consolidate the pos calculation by moving it outside the if/else

packages/atclient/src/metadata.c
```diff
@@ -1210,112 +1211,112 @@ int atclient_atkey_metadata_to_protocol_str(const atclient_atkey_metadata *metad
...
-  if (strlen(*metadata_str) != (expected_metadatastr_len)) {
+
+  if (strlen(*metadata_str) != expected_metadatastr_len) {
     ret = 1;
-    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "metadata_str length mismatch: %lu != %lu\n", strlen(*metadata_str),
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR,
+                 "metadata_str length mismatch: %lu != %lu\nmetadata:", strlen(*metadata_str),
                  (expected_metadatastr_len));
+
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "metadata_str: %s\n", (*metadata_str));
```
xavierchanth, line 1319: Added some more logging in case this comes up again

gkc (approved):
LGTM

PR #412, opened 2024-09-27
feat: add apkam symmetric key to atkeys and atkeys_file

xavierchanth:
**- What I did**

**- How I did it**

**- How to verify it**

feat: add apkam symmetric key to atkeys and atkeys_file

### DIFF
--- a/packages/atclient/include/atclient/atkeys.h
+++ b/packages/atclient/include/atclient/atkeys.h
@@ -14,6 +14,7 @@
 #define ATCLIENT_ATKEYS_ENCRYPT_PRIVATE_KEY_INDEX 0
 #define ATCLIENT_ATKEYS_SELF_ENCRYPTION_KEY_INDEX 0
 #define ATCLIENT_ATKEYS_ENROLLMENT_ID_INDEX 0
+#define ATCLIENT_ATKEYS_APKAM_SYMMETRIC_KEY_INDEX 0
 
 #define ATCLIENT_ATKEYS_PKAM_PUBLIC_KEY_INITIALIZED (VALUE_INITIALIZED << 0)
 #define ATCLIENT_ATKEYS_PKAM_PRIVATE_KEY_INITIALIZED (VALUE_INITIALIZED << 1)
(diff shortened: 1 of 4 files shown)


xavierchanth:
Setting this to draft, since I haven't updated tests & examples, @srieteja can you try this on Monday and see if it solves your problems?

You can probably cherry pick these commits onto your working branch to try the changes out.

packages/atclient/include/atclient/atkeys.h
```diff
@@ -48,6 +50,7 @@ typedef struct atclient_atkeys {
 
   char *self_encryption_key_base64; // base64 encoded, AES-256 key, decrypted
 
+  char *apkam_symmetric_key_base64;
```
JeremyTubongbanua, line 53: if this is going to be the 5th thing inside of the struct, we should switch around so that

instead of this:
```h
#define ATCLIENT_ATKEYS_ENROLLMENT_ID_INITIALIZED (VALUE_INITIALIZED << 5)
#define ATCLIENT_ATKEYS_APKAM_SYMMETRIC_KEY_INITIALIZED (VALUE_INITIALIZED << 6)
```

we do this instead
```
#define ATCLIENT_ATKEYS_APKAM_SYMMETRIC_KEY_INITIALIZED (VALUE_INITIALIZED << 5)
#define ATCLIENT_ATKEYS_ENROLLMENT_ID_INITIALIZED (VALUE_INITIALIZED << 6)
```

so that we keep the order of things consistent

packages/atclient/include/atclient/atkeys.h
```diff
@@ -82,7 +85,11 @@ int atclient_atkeys_set_encrypt_private_key_base64(atclient_atkeys *atkeys, cons
 int atclient_atkeys_set_self_encryption_key_base64(atclient_atkeys *atkeys, const char *selfencryptionkeybase64,
                                                    const size_t selfencryptionkeybase64len);
 
-int atclient_atkeys_set_enrollment_id(atclient_atkeys *atkeys, const char *enrollment_id, const size_t enrollment_id_len);
+int atclient_atkeys_set_apkam_symmetric_key_base64(atclient_atkeys *atkeys, const char *apkamsymmetrickeybase64,
+                                                   const size_t apkamsymmetrickeybase64len);
```
JeremyTubongbanua, line 89: Just a nit

`apkam_symmetric_key_base64` is the name of the struct, so the parameter name should be the same

JeremyTubongbanua (approved):

packages/atclient/include/atclient/atkeys.h
xavierchanth: Will do

packages/atclient/include/atclient/atkeys.h
xavierchanth: All the other signatures follow this naming convention though

PR #417, opened 2024-10-03
feat: changes in `examples/desktop/` and `atclient_connection_read`

JeremyTubongbanua:
**- What I did**
- Removed logging of atserver_host and atserver_port which is a duplicate log of the DEBUG response after requesting atserver host/port from atDirectory
- Removed a debug log in `atclient_connection_read`

I think these two logs were used during implementation but realistically don't need to be there. The first set of logs I removed was ensuring that the atserver_host:atserver_port were properly parsed, but that implementation has always been perfect. The second set of logs is also another set of logs that were used during implementation and debugging and can be removed. They aren't telling us any new information of what's going on behind the scenes, so it's just extra noise that I think we can remove safely.

- Fixed reallocating in `atclient_connection_read`

- Added `examples/playground` which is a useful template for debugging application code while using the atSDK as a dependency
- Made refactor changes to `examples/repl`

**- How I did it**
- Most of this PR is examples/ changes that I find useful for debugging

feat: misc changes in examples/desktop


examples/desktop/repl/3rdparty/argparse/include/argparse/LICENSE
xavierchanth, line 1: We shouldn't relicense argparse. The LICENSE in the argparse files & the README suffice

xavierchanth (changes requested):
Major thing:

- Don't relicense argparse

Nit:

- Use a better naming convention than `main*.c` in the playground or at least put some comments in the tops of the files
  - Secondarily to this, I think it's fine to use a GLOB for the files in playground if the files have better names

JeremyTubongbanua:
> Major thing:
> 
> * Don't relicense argparse
> 
> Nit:
> 
> * Use a better naming convention than `main*.c` in the playground or at least put some comments in the tops of the files
>   
>   * Secondarily to this, I think it's fine to use a GLOB for the files in playground if the files have better names

deleted license - https://github.com/atsign-foundation/at_c/pull/417/commits/00d769f87a67a8d97cf64deeb8eafbee6966f1ca

made it so we only have one main.c in playground (and added `tools/` to playground) - https://github.com/atsign-foundation/at_c/pull/417/commits/8c1210639a605910126b482b87d15a8ed7ef8378

xavierchanth (approved):

PR #419, opened 2024-10-04
feat: atchops_rsa_key_generate_base64

JeremyTubongbanua:
**- What I did**
- Added a new atchops_rsa function: `atchops_rsa_key_generate_base64`
- Refactored `atchops_rsa_key_generate` to utilize the new `atchops_rsa_key_generate_base64` function
- Added a test to ensure that `atchops_rsa_key_generate_base64`: 1. properly generated a key, 2. the keys could be used to populate our rsa key srtucts and 3. the keys could be used to encrypt/decrypt a basic plaintext
- Refactored `test_rsa_generate` to `test_rsa_key_generate`

**- How to verify it**
- New test passes
- and previous tests pass

feat: atchops_rsa_key_generate_base64


xavierchanth:
Can you explain a little more about history of aes generate vs history of rsa generate, we had an `aes_generate_base64` and it was removed to keep the public interface simple, so why are we adding `rsa_generate_base64`

JeremyTubongbanua:
To put it simply, atchops is a wrapper around MbedTLS (we make it simple for us to call MbedTLS for our use case)

MbedTLS AES generation: outputs bytes
MbedTLS RSA generation: outputs base64

So therefore

Atchops AES generation: outputs bytes
Atchops RSA generation: outputs base64

xavierchanth (approved):

PR #431, opened 2024-10-18
chore: improved monitor read, and TODOs for everything else

xavierchanth:
**- What I did**

- Modified the monitor read to handle cases where we should retry reading instead of returning an error.
- Added TODOs all over the code base where this same thing should be handled for read and writes
- Will open an accompanying issue

**- How I did it**

**- How to verify it**

chore: improved monitor read, and TODOs for everything else

### DIFF
--- a/packages/atclient/src/monitor.c
+++ b/packages/atclient/src/monitor.c
@@ -128,10 +128,35 @@ int atclient_monitor_read(atclient *monitor_conn, atclient *atclient, atclient_m
     }
 
     size_t off = chunksize * chunks;
-    for (int i = 0; i < chunksize; i++) {
+    int i = 0;
+    while (i < chunksize) {
       ret = mbedtls_ssl_read(&(monitor_conn->atserver_connection.ssl), (unsigned char *)buffer + off + i, 1);
-      if (ret <= 0 || buffer[off + i] == '\n') {
+      // successfully read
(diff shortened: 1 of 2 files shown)


packages/atclient/src/monitor.c
```diff
@@ -128,13 +128,38 @@ int atclient_monitor_read(atclient *monitor_conn, atclient *atclient, atclient_m
...
+      }
+
+      // Handle errors
+      switch (ret) {
+        // Special error cases where we should try reading again
+      case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:  // async operation in progress
+      case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS: // crypto operation in progress
+      case MBEDTLS_ERR_SSL_WANT_READ:          // handshake incomplete
+      case MBEDTLS_ERR_SSL_WANT_WRITE:         // handshake incomplete
+        usleep(10000);                         // Try again in 10 milliseconds
```
JeremyTubongbanua, line 153: This is `10 seconds` not `10 milliseconds`

JeremyTubongbanua (approved):

packages/atclient/src/monitor.c
xavierchanth: sleep is in seconds, usleep is in microseconds. *1000 for milliseconds.

PR #438, opened 2024-10-30
feat: virtual environments in functional tests

### LINKED ISSUE #211: Use virtual environment for functional tests rather than production atsigns

JeremyTubongbanua:
closes #211 

**- What I did**
WIP

**- How I did it**

**- How to verify it**

feat: virtual environments in functional tests


JeremyTubongbanua:
I'm close to refactoring the functional tests to work with the virtual environment

Tests just aren't passing right now

PR #439, opened 2024-11-04
fix: use sized types for more consistent cross-platform support

xavierchanth:
**- What I did**

- Replaced occurences of int, long, unsigned long with uint64_t and int64_t.
  - Some of these are unlikely to overflow, but it doesn't hurt, since dart uses 64 bit integers, and these values are exchanged with the atServer.

**- How I did it**

**- How to verify it**

- When I make the same changes to NoPorts, I will be able to verify

fix: use sized types for more consistent cross-platform support

### DIFF
--- a/packages/atclient/include/atclient/version.h
+++ b/packages/atclient/include/atclient/version.h
@@ -1,6 +1,6 @@
 #ifndef ATCLIENT_VERSION_H
 #define ATCLIENT_VERSION_H
 
-#define ATCLIENT_ATSDK_VERSION "0.1.0"
+#define ATCLIENT_ATSDK_VERSION "0.3.0"
 
 #endif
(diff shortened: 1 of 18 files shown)


xavierchanth:
Oops, more commits coming for the tests... clangd doesn't index the tests, only include and src.

packages/atclient/include/atclient/version.h
```diff
@@ -1,6 +1,6 @@
 #ifndef ATCLIENT_VERSION_H
 #define ATCLIENT_VERSION_H
 
-#define ATCLIENT_ATSDK_VERSION "0.0.1"
+#define ATCLIENT_ATSDK_VERSION "0.1.0"
```
cpswan: Should this match the version in the change log?

xavierchanth: Yes... this was a situation where I bumped the version, then went to the releases to check what version we were actually on. Then went to add everything to the change log and forgot to come back.

cpswan (approved):

PR #446, opened 2024-11-11
ci: stricter tests

xavierchanth:
**- What I did**

More diagnostic information in our build source ci:

- Print version info for GCC
- Print CMake cache variables

Stricter build flags in our build source ci:

- Use gcc as the compiler with stricter flags so that we are catching more
  problems than we currently are.

**- How I did it**

**- How to verify it**

ci: stricter tests

### DIFF
--- a/.github/workflows/build_source_tests.yaml
+++ b/.github/workflows/build_source_tests.yaml
@@ -16,7 +16,9 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Install atsdk
-        run: cmake -S . -B build -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS="-Wall -Wextra -Werror-implicit-function-declaration"
+        run: cmake -S . -B build -DCMAKE_C_COMPILER=gcc \
+          -DCMAKE_C_FLAGS="-Wall -Wextra -Werror-implicit-function-declaration" \
+          -DCMAKE_COMPILE_WARNING_AS_ERROR=true
 
       - name: Build atsdk
         run: sudo cmake --build build --target install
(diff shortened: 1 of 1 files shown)


xavierchanth:
The NoPorts e2e tests run on an external vm but they are using the same version of gcc...

```
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

It also runs a very similar command to what we now run here:

```
    local base_dir="./sshnpd"
    local build_dir="$base_dir/build"
    mkdir -p "$build_dir"
    cmake -B $build_dir -S $base_dir -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS="-Wall -Wextra -Werror-implicit-function-declaration" -DBUILD_TESTS=off
    cmake --build $build_dir
```

xavierchanth:
at_c (as of now):
```
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```


```
      - name: Install atsdk
        run: cmake -S . -B build -DCMAKE_C_COMPILER=gcc \
          -DCMAKE_C_FLAGS="-Wall -Wextra -Werror-implicit-function-declaration" \
          -DCMAKE_COMPILE_WARNING_AS_ERROR="on"
```

xavierchanth:
I'm truly at a loss as to why one is suceeding and the other is not... I'm going to see if I can retrieve (and later) purge the CMake cache on the NoPorts test runner and see if any lingering rules solve the problem...

xavierchanth:
I did a diff of the two cmake caches and the only things different were:
- the machine specific path to the deps directory
- fetch content and build test variables that are specific to dependencies only used in NoPorts

cpswan (approved):

PR #447, opened 2024-11-12
chore: wrap all .h files with extern "C" macros

xavierchanth:
**- What I did**

- Added extern C macros to all header files in the project.

**- How I did it**

**- How to verify it**

- **Enable ignore whitespace for the review**
  > a bunch of those files were not following our standard formatting rules

- I checked to see that the macros were well formed, but I couldn't find any
  clear information as to what is allowed inside them... I'm not sure how the
  linker resolves includes and defines inside the extern block, nor do I have a
  C++ project to test.

chore: wrap all .h files with extern "C" macros

### DIFF
--- a/examples/desktop/repl/include/repl/args.h
+++ b/examples/desktop/repl/include/repl/args.h
@@ -1,5 +1,8 @@
 #ifndef REPL_ARGS_H
 #define REPL_ARGS_H
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 #include <stdint.h>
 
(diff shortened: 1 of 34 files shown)


cpswan:
Can you please add a little context on why?

xavierchanth:
> Can you please add a little context on why?

@cconstab is working on a demo which requires c++. This should allow him to link our c sdk into that project.

cpswan (approved):

xavierchanth:
I would like to run #446 against these changes.

PR #456, opened 2024-12-04
chore: Simpler deps

xavierchanth:
**- What I did**

- Removed uuid4 as an external dependency
  - It is now included statically as private static functions so it can be easily modified to add other platforms
- json header removed from atclient
  - json now provided by atcommons
  - wrapper around json implementation to make it platform specific 

**- How I did it**

**- How to verify it**


CMakeLists.txt
```diff
@@ -45,47 +45,46 @@ message(STATUS "Building atauth")
...
+    mbedcrypto
+    everest
+    p256m
+    atchops
+    cjson
+    atlogger
+    atclient
+    atcommons
+    atauth
+  )
```
xavierchanth, line 63: removed uuid4-static

packages/atauth/CMakeLists.txt
```diff
@@ -20,131 +21,134 @@ cmake_policy(SET CMP0135 NEW)
...
+    mbedtls
+    mbedx509
+    mbedcrypto
+    everest
+    p256m
+    atclient
+    atcommons
+    atlogger
+    atchops
+  )
```
xavierchanth, line 99: removed uuid4-static

packages/atchops/CMakeLists.txt
```diff
@@ -94,14 +89,7 @@ if(NOT ESP_PLATFORM)
...
 
   # Link dependencies to library targets
-  set(
-    ATCHOPS_INSTALL_TARGETS
-    atlogger
-    mbedcrypto
-    everest
-    p256m
-    uuid4-static
-  )
```
xavierchanth, line 104: removed uuid4-static

packages/atchops/src/uuid.c
```diff
@@ -1,11 +1,36 @@
...
+#include <stdint.h>
+#include <stdio.h>
 #include <string.h>
-#include <uuid4/uuid4.h>
 
 #define TAG "uuid"
 
+// uuid4 was originally an external library which was depended upon as
+// an external dependency, it has been moved into this file as private
+// symbols so that it's easier to make platform specific overrides
```
xavierchanth, line 12: Added the code from our fork of uuid4. So that it's easier to make platform specific changes, they can now happen all in one repo.

packages/atclient/include/atclient/atnotification.h
```diff
@@ -4,7 +4,7 @@
 extern "C" {
 #endif
 
-#include "atclient/cjson.h"
+#include "atcommons/json.h"
```
xavierchanth, line 7: moved json header to only be in atcommons, atclient now depends on atcommons.

Also renamed from cjson to json to be generic to any json library.

packages/atclient/include/atclient/atnotification.h
```diff
@@ -96,7 +96,6 @@ void atclient_atnotification_init(atclient_atnotification *notification);
 void atclient_atnotification_free(atclient_atnotification *notification);
 
 int atclient_atnotification_from_json_str(atclient_atnotification *notification, const char *json_str);
-int atclient_atnotification_from_cjson_node(atclient_atnotification *notification, const cJSON *root);
```
xavierchanth, line 99: removed cjson dependent function from public api

packages/atclient/include/atclient/metadata.h
```diff
@@ -286,16 +286,6 @@ int atclient_atkey_metadata_clone(atclient_atkey_metadata *dst, const atclient_a
...
-/**
- * @brief Populates the metadata struct from a cJSON pointer. This function is good for debugging.
- *
- * @param metadata the metadata struct to populate
- * @param json the json object to populate from
- * @return int 0 on success
- */
-
-int atclient_atkey_metadata_from_cjson_node(atclient_atkey_metadata *metadata, const cJSON *json);
-
```
xavierchanth, line 298: removed cjson dependent function from public api

packages/atclient/src/atnotification.c
```diff
@@ -5,6 +5,11 @@
 
 #define TAG "atnotification"
 
+// Json library specific
+#if ATCLIENT_JSON_PROVIDER == ATCLIENT_JSON_PROVIDER_CJSON
+static int atclient_atnotification_from_cjson_node(atclient_atnotification *notification, const cJSON *root);
+#endif
+
```
xavierchanth, line 12: This function is private now, and conditionally linked into the program only when cJSON is available

packages/atclient/src/atnotification.c
```diff
@@ -84,340 +89,165 @@ void atclient_atnotification_free(atclient_atnotification *notification) {
   }
 }
 
-int atclient_atnotification_from_json_str(atclient_atnotification *notification, const char *json_str) {
```
xavierchanth, line 87: The diff in this file is really messed up because it was misformatted

I've taken the json-library specific functions and moved them to the bottom of the file

packages/atclient/src/atnotification.c
```diff
@@ -1907,3 +1579,340 @@ int atclient_atnotification_set_decrypted_value(atclient_atnotification *notific
   goto exit;
 exit: { return ret; }
 }
+
```
xavierchanth, line 1582: Here's the new placement for the functions. We will need to implement `atclient_atnotification_from_json_str` for each json library we want to support.

packages/atclient/src/metadata.c
```diff
@@ -11,6 +11,11 @@
 
 #define TAG "metadata"
 
+// Json library specific
+#if ATCLIENT_JSON_PROVIDER == ATCLIENT_JSON_PROVIDER_CJSON
+static int atclient_atkey_metadata_from_cjson_node(atclient_atkey_metadata *metadata, const cJSON *json);
+#endif
+
```
xavierchanth, line 18: Exact same story as with atnotification.

packages/atcommons/include/atcommons/cjson.h
```diff
@@ -1,10 +0,0 @@
-#ifndef ATCOMMONS_CJSON_H
-#define ATCOMMONS_CJSON_H
-
-#if defined(CONFIG_IDF_TARGET_ESP32)
-#include <cjson.h>
-#else
-#include "cJSON.h"
-#endif
-
-#endif
```
xavierchanth, line 10: moved to atcommons/json.h

packages/atcommons/include/atcommons/json.h
```diff
@@ -1,9 +1,12 @@
-#ifndef ATCLIENT_CJSON_H
-#define ATCLIENT_CJSON_H
+#ifndef ATCLIENT_JSON_H
+#define ATCLIENT_JSON_H
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define ATCLIENT_JSON_PROVIDER_CJSON 1
+#define ATCLIENT_JSON_PROVIDER ATCLIENT_JSON_PROVIDER_CJSON
```
xavierchanth, line 8: default json provider

packages/atchops/src/uuid.c
JeremyTubongbanua: Can we add links like

"The original uuid4 repository by rxi: https://github.com/rxi/uuid4"
"Atsign's fork of uuid4 by rxi: https://github.com/atsign-foundation/uuid4"

JeremyTubongbanua (dismissed):

JeremyTubongbanua:
A question I have is:

If we're getting rid of our uuid4 fork and just having the code in our codebase (which is okay because of open LICENSING), why don't we do the same thing for cjson so we don't have to worry about managing the ATCLIENT_JSON_PROVIDER constant?

packages/atchops/src/uuid.c
xavierchanth: I've got the license in, but sure, I'll add the link

xavierchanth:
> A question I have is:
> 
> If we're getting rid of our uuid4 fork and just having the code in our codebase (which is okay because of open LICENSING), why don't we do the same thing for cjson so we don't have to worry about managing the ATCLIENT_JSON_PROVIDER constant?

We could, but I've focused on doing the minium viable work for gettings things up on the portenta. I actually have a different non-cmake implementation of cJSON which has been ported to Arduino which will be included in the Arduino library folder.

Also, cJSON and mbedtls are libraries that tend to be available in more places. So I think it's ok to statically link against cmake fetches for now, leaving the option open to dynamically link if needed.

xavierchanth:
I should add, I don't see a world where we ever _need_ to dynamically link uuid4, it's a tiny library, even in comparison to cJSON (which is already quite small)

JeremyTubongbanua:
> > A question I have is:
> > If we're getting rid of our uuid4 fork and just having the code in our codebase (which is okay because of open LICENSING), why don't we do the same thing for cjson so we don't have to worry about managing the ATCLIENT_JSON_PROVIDER constant?
> 
> We could, but I've focused on doing the minium viable work for gettings things up on the portenta. I actually have a different non-cmake implementation of cJSON which has been ported to Arduino which will be included in the Arduino library folder.
> 
> Also, cJSON and mbedtls are libraries that tend to be available in more places. So I think it's ok to statically link against cmake fetches for now, leaving the option open to dynamically link if needed.

The option to dynamically link makes more sense to me

I didn't see uuid4 on ESP-IDF, so it makes sense to just have the code live directly in our repo.

JeremyTubongbanua (dismissed):

xavierchanth:
Ah... my merge resolution commit is still on local

xavierchanth:
Ok, conflicts have been resolved and everything is linking correctly. Just waiting on the tests to double check everything.

JeremyTubongbanua (approved):

PR #461, opened 2024-12-05
feat: Partial Arduino support

xavierchanth:
**- What I did**

### Implemented backwards compatibility with MbedTLS 2

By adding a define statement which is automatically enabled when working with
certain platforms. Later versions of mbedtls 2 have already deprecated all the
additional parameters  that are conditionally passed when working with mbedtls 2.
(e.g. specifying which key to use with rsa encrypt/decrypt/sign/verify - these
operations already have a specific key which should be used)

### Added a generators/arduino directory which generates the arduino library
- There are template files which get pulled into the final library build
  - File templates are postfixed with `template` to prevent the arduino compiler
    from trying to build them
- Added platform.h to atchops which resolves platform specifics for
  shared dependencies between atchops and atclient
  Most headers now include platform.h (some, like constants.h don't)
- Changed atcommons/json.h's define to follow a similar format to platform.h
- The generator has two scripts:
  - generate.sh - (re)generate the arduino library
  - install.sh - prompt for permission to delete the old & copy over the newly
    generated library into the Arduino IDE library folder.
    Arduino has a specific folder for IDE projects, and it will share libraries
    between those projects, only compiling if that library is included in the project
- This build currently is only enabled for `TARGET_PORTENTA_H7` but in theory
  should support any Arduino board/core which includes mbedtls2 in the core.
  We would just need to enable the correct defines for those other targets

Note:
- atchops/platform.h provides information about dependencies shared between:
  - atchops
  - atclient
- atcommons/json.h provides information about dependencies shared between:
  - atcommons
  - atclient
This may seem weird, but atcommons and atchops have no dependency on eachother.
So instead of introducing a new library just to provide a single header file,
both files have a `NOTE:` comment referencing eachother.

### atlogger changes

Removed "ENABLE_TIMESTAMPS" bit flag from opts in atlogger.
To disable timestamps, the code must define:

`#define ATLOGGER_DISABLE_TIMESTAMPS` before including atlogger.h
This means that atlogger will disable timestamps at build time which
will prevent linker errors if the clock_gettime symbol is not found.

I've kept in the `atlogger_set_opts` function (which for now does nothing)
to minimize breaking changes.

### Coming in a later PR

- Documentation on how to generate and use the library
- Example sketchbook of atclient setup & sending notifications
- Wrapper around the Arduino TCP socket library as mbedtls's net_socket.h
  library is disabled
- Any other changes required once I sort out networking

**- How I did it**

- Lots of patience

**- How to verify it**

feat: partial arduino support


xavierchanth:
Once #456 has been merged, the base of this branch should automatically point itself to trunk. At least that was something I thought I saw GitHub do, and wanted to see if it was the case.

xavierchanth:
> Once #456 has been merged, the base of this branch should automatically point itself to trunk. At least that was something I thought I saw GitHub do, and wanted to see if it was the case.

Nope, must have been GitButler's stacked PR feature doing it.

generators/arduino/atsdk/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+cmake_minimum_required(VERSION 2.27)
```
cpswan: What drove the nudge up to 2.27?

xavierchanth: I think a mixup, that file is no longer needed actually. I will drop it from the PR. Previously I was using cmake for external dependencies, but everything is statically available now

cpswan (approved):

PR #464, opened 2024-12-07
feat: Arduino atlogger

xavierchanth:
**- What I did**

### Arduino overrides

- Moved the .*template files to be normal files in a lib/ folder, Arduino doesn't compile these files!
  - Reverted the neovim & vscode settings needed for these templates.
  - I will be adding more of these files to handle networking implementation, so we really needed a better way before I made that effort.
- Removed references to portenta from our main packages
  - Part of the Arduino library generation now adds defines and includes to the appropriate headers

### Testing

This originally wasn't intended be such a large change, sorry, it should have been a separate PR.

- Allow the test atsigns to be overridden in cmake:
```sh
cmake -B $repo_base/build -S $repo_base \
  -DCMAKE_INSTALL_PREFIX="$HOME/.local/" \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_C_COMPILER=gcc \
  -DCMAKE_C_FLAGS="-std=c99 -Werror" \
  -DATSDK_BUILD_TESTS=ON \
  -DFIRST_ATSIGN='"@first"' \
  -DSECOND_ATSIGN='"@second"'
```

> The inner quotes are necessary, otherwise the define statement will be processed as:
`#define FIRST @first` instead of `#define FIRST "@first"`

- Fixed build errors in tests (casting issues, and format specifiers)
- Split atSDK level tests into two modes: unit test and functional test.
- By default `ATSDK_BUILD_TESTS` now builds all tests, and there are also:
  `ATSDK_BUILD_FUNCTIONAL_TESTS` and `ATSDK_BUILD_UNIT_TESTS` which allow us
  split the tests into two parallel runs for CI
  - Also conditionally `enable_testing()` in the root cmake so the entire test
    suite can be run with a single ctest command.
    ```sh
    ctest --test-dir "$repo_base/build" \
      --output-on-failure \
      --parallel 5
    ```
- Raised test timeouts to reduce flakes
  - unit : 2 -> 5 seconds
  - functional : 60 -> 90 seconds

**- How I did it**

**- How to verify it**

feat: Arduino atlogger


xavierchanth:
Sorry for the email spam, some changes I made to the functional tests were working on local as I had atsdk exported, but that isn't the case for the test runner.

cpswan (approved):

PR #475, opened 2024-12-13
feat: move mbedtls sockets into its own layer under connection

xavierchanth:
**- What I did**

- Moved TLS sockets into a generic interface
  - Connection responsible for managing connections to atDirectory/atServer and parsing/stripping prompts from those connections
  - TLS sockets responsible for establishing TLS connections, and reading/writing raw bytes to and from those sockets

- TLS socket implementation can be swapped out without the need to reimplement all of connection which also contains a bunch of parsing logic

**- How I did it**

**- How to verify it**

feat: move mbedtls sockets into its own layer under connection

### DIFF
--- a/packages/atclient/include/atclient/connection.h
+++ b/packages/atclient/include/atclient/connection.h
@@ -1,7 +1,7 @@
 /*
  *
  * The connection family of types and methods represents a single connection to
- * if you want a pure socket representation see net_socket.h.
+ * if you want a pure socket representation see socket.h.
  *
  * At the moment _socket represents a singular tcp socket, but in the future it may be altered
  * to be a union of different connection types, such as a websocket or other construct.
--- a/packages/atclient/include/atclient/connection.h
+++ b/packages/atclient/include/atclient/connection.h
@@ -1,12 +1,24 @@
+/*
+ *
+ * The connection family of types and methods represents a single connection to
+ * if you want a pure socket representation see net_socket.h.
+ *
+ * At the moment _socket represents a singular tcp socket, but in the future it may be altered
+ * to be a union of different connection types, such as a websocket or other construct.
+ * It is considered an internal construct and is subject to breaking changes across
+ * minor releases, especially while at_c remains in beta status.
+ *
+ */
 #ifndef ATCLIENT_CONNECTION_H
 #define ATCLIENT_CONNECTION_H
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#include "atchops/mbedtls.h"
-#include "atclient/connection_hooks.h"
 #include <atchops/platform.h> // IWYU pragma: keep
+
+#include "atclient/connection_hooks.h"
+#include "atclient/socket.h"
 #include <stdbool.h>
 #include <stddef.h>
 
@@ -31,12 +43,8 @@ typedef struct atclient_connection {
   // _is_connection_enabled also serves as an internal boolean to check if the following mbedlts contexts have been
   // initialized and need to be freed at the end
   bool _is_connection_enabled : 1;
-  mbedtls_net_context net;
-  mbedtls_ssl_context ssl;
-  mbedtls_ssl_config ssl_config;
-  mbedtls_x509_crt cacert;
-  mbedtls_entropy_context entropy;
-  mbedtls_ctr_drbg_context ctr_drbg;
+
+  struct atclient_tls_socket _socket;
 
   bool _is_hooks_enabled : 1;
   atclient_connection_hooks *hooks;
@@ -80,8 +88,7 @@ int atclient_connection_connect(atclient_connection *ctx, const char *host, cons
  * @param value_max_len the maximum length of the data to read, setting this to 0 means no limit
  * @return int 0 on success
  */
-int atclient_connection_read(atclient_connection *ctx, unsigned char **value, size_t *value_len,
-                             const size_t value_max_len);
+int atclient_connection_read(atclient_connection *ctx, unsigned char **value, size_t *value_len);
 
 /**
  * @brief Write data to the connection
--- a/packages/atclient/include/atclient/connection.h
+++ b/packages/atclient/include/atclient/connection.h
@@ -30,23 +30,21 @@ typedef enum atclient_connection_type {
 
 typedef struct atclient_connection {
   atclient_connection_type type; // set in atclient_connection_init
-
+  uint16_t port;                 // example: 64
   bool _is_host_initialized : 1;
-  char *host; // example: "root.atsign.org"
-
   bool _is_port_initialized : 1;
-  uint16_t port; // example: 64
+  bool _is_connection_enabled : 1;
+  bool _is_hooks_enabled : 1;
+
+  char *host; // example: "root.atsign.org"
 
   // atclient_connection_connect sets this to true and atclient_connection_disconnect sets this to false
   // this does not mean that the connection is still alive, it just means that the connection was established at least
   // once, at some  point, check atclient_connection_is_connected for a live status on the connection
   // _is_connection_enabled also serves as an internal boolean to check if the following mbedlts contexts have been
   // initialized and need to be freed at the end
-  bool _is_connection_enabled : 1;
 
   struct atclient_tls_socket _socket;
-
-  bool _is_hooks_enabled : 1;
   atclient_connection_hooks *hooks;
 } atclient_connection;
 
--- a/packages/atclient/include/atclient/monitor.h
+++ b/packages/atclient/include/atclient/monitor.h
@@ -9,6 +9,11 @@ extern "C" {
 #include <atchops/platform.h> // IWYU pragma: keep
 #include <stdbool.h>
 
+// HACK let's just get it working for now this is so wrong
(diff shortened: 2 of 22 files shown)


packages/atclient/include/atclient/socket.h
```diff
@@ -0,0 +1,113 @@
...
+ * @param socket The socket structure to initialize
+ */
+void atclient_raw_socket_init(struct atclient_raw_socket *socket);
+
+/**
+ * @brief Frees resources associated with a network socket
+ *
+ * @param socket The socket structure to free resources from
+ */
+void atclient_raw_socket_free(struct atclient_raw_socket *socket);
```
xavierchanth, line 50: Raw sockets only have init and free for now. I plan to add more functionality later, but all we need is init and free right now, as the ssl context takes ownership of the object.

packages/atclient/include/atclient/socket_mbedtls.h
```diff
@@ -0,0 +1,32 @@
...
+#include <mbedtls/ctr_drbg.h>
+#include <mbedtls/entropy.h>
+#include <mbedtls/net_sockets.h>
+#include <mbedtls/ssl.h>
+#include <mbedtls/threading.h>
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Make this type more portable to consume later
```
xavierchanth: TODO: remove lingering comment

packages/atclient/include/atclient/socket_shared.h
```diff
@@ -0,0 +1,57 @@
+// IWYU pragma: private, include "atclient/net_socket.h"
+// IWYU pragma: friend "net_socket_mbedtls.*"
```
xavierchanth: TODO replace instances of net_socket with the updated file names

packages/atclient/src/atclient.c
```diff
@@ -261,7 +268,6 @@ int atclient_pkam_authenticate(atclient *ctx, const char *atsign, const atclient
   /*
    * 5. Get atserver_host and atserver_port
    */
-  bool should_free_atserver_host;
```
xavierchanth, line 264: Moved up, as it was undefined in some scenarios due to goto statements, there are a couple of other instances I caught.

packages/atclient/src/atclient_get_atkeys.c
```diff
@@ -46,7 +46,7 @@ int atclient_get_atkeys(atclient *atclient, atclient_atkey **atkey, size_t *outp
 
   char scan_cmd[scan_cmd_size];
 
-  const size_t recv_size = 8192; // TODO change using atclient_connection_read which will handle realloc
+  const size_t recv_size = 16384; // TODO change using atclient_connection_read which will handle realloc
```
xavierchanth, line 49: Increased this because scan results can be large, the TLS socket now creates a heap allocated buffer, but read and send still expect pre-allocated buffers. I didn't have time to uptake this in existing code yet. If the buffer in the heap is larger than the preallocated size, then the functions log and return an error code.

packages/atclient/src/connection.c
xavierchanth, line 1: All of the removed code is now located in sockets layer, instead  we call socket layer here.

packages/atclient/src/atclient.c
JeremyTubongbanua: Good catch

JeremyTubongbanua (dismissed):

packages/atclient/src/atclient.c
xavierchanth: clangd caught it, not sure if it's the new llvm update or the stricter rules I added to my settings. Good tooling ftw

packages/atclient/include/atclient/connection.h
```diff
@@ -1,12 +1,24 @@
...
+ */
 #ifndef ATCLIENT_CONNECTION_H
 #define ATCLIENT_CONNECTION_H
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#include "atchops/mbedtls.h"
-#include "atclient/connection_hooks.h"
 #include <atchops/platform.h> // IWYU pragma: keep
```
xavierchanth, line 18: platform needs to be first, as socket depends on a value in it.

packages/atclient/include/atclient/connection.h
```diff
@@ -18,27 +30,21 @@ typedef enum atclient_connection_type {
...
+  uint16_t port;                 // example: 64
   bool _is_host_initialized : 1;
-  char *host; // example: "root.atsign.org"
-
   bool _is_port_initialized : 1;
-  uint16_t port; // example: 64
+  bool _is_connection_enabled : 1;
+  bool _is_hooks_enabled : 1;
+
+  char *host; // example: "root.atsign.org"
```
xavierchanth, line 39: Grouped the booleans in this struct to marginally improve the alignment. Since the alignment is quite high, should help reduce the chances of cache misses

packages/atclient/include/atclient/monitor.h
```diff
@@ -9,6 +9,11 @@ extern "C" {
 #include <atchops/platform.h> // IWYU pragma: keep
 #include <stdbool.h>
 
+// HACK let's just get it working for now this is so wrong
```
cpswan: Why -37?
What's the proper way to do this?

packages/atclient/include/atclient/socket_mbedtls.h
cpswan: It might be worth putting the TODO in the comment (so it's in the code rather than what will be a closed PR comment)

packages/atclient/include/atclient/socket_shared.h
cpswan: When is this TODO expected to be addressed?

tests/functional_tests/tests/test_atclient_get_atkeys.c
```diff
@@ -146,6 +146,7 @@ exit: {
 
 static int test_4_atclient_get_atkeys_null_regex(atclient *ctx, const char *scan_regex, const bool showhidden) {
   int ret = 1;
+  return 0; // NOTE: disabled temporarily DONT LET THIS GET MERGED
```
cpswan: I'll make PR draft until this is addressed

cpswan (changes requested):

packages/atclient/include/atclient/socket_mbedtls.h
xavierchanth: > It might be worth putting the TODO in the comment (so it's in the code rather than what will be a closed PR comment)

I normally would, but I the TODO is to remove the comment because it's no longer relevant

packages/atclient/include/atclient/monitor.h
xavierchanth: I was trying to avoid leaking Arduino implementation details into the main package, but I've deemed it unpractical. I've added a new macro for `ATCLIENT_SSL_TIMEOUT_EXITCODE`

Which provides:
- MBEDTLS_ERR_SSL_TIMEOUT with mbedtls
- `-101` for Arduino with bearssl (typically -1 is the only negative number used by their libraries)
- A preprocessor otherwise a preprocessor error and it must be defined externally

xavierchanth: We just need this number to be unique based to the other error types which may get returned by whatever implementation of tls sockets are being used

packages/atclient/include/atclient/socket_mbedtls.h
xavierchanth: Addressed in [40f6221](https://github.com/atsign-foundation/at_c/pull/475/commits/40f6221f743d7536dd9dd3fa102657fb0bc8230f)

packages/atclient/include/atclient/socket_shared.h
xavierchanth: Addressed in [40f6221](https://github.com/atsign-foundation/at_c/pull/475/commits/40f6221f743d7536dd9dd3fa102657fb0bc8230f)

packages/atclient/include/atclient/monitor.h
xavierchanth: Addressed in [40f6221](https://github.com/atsign-foundation/at_c/pull/475/commits/40f6221f743d7536dd9dd3fa102657fb0bc8230f)

tests/functional_tests/tests/test_atclient_get_atkeys.c
xavierchanth: Addressed in [40f6221](https://github.com/atsign-foundation/at_c/pull/475/commits/40f6221f743d7536dd9dd3fa102657fb0bc8230f)

xavierchanth:
@cpswan it's worth discussing how we want to test this. It is tested through functional tests with notify, monitor, connection, pkam_authenticate & get_atkeys. But we now have the ability to test the pure tls socket client, we would need to create a test harness though.

cpswan (dismissed):

cpswan:
@XavierChanth agree that more tests would be good, though maybe open a fresh issue for those.

cpswan (approved):

PR #479, opened 2024-12-19
feat: let the caller handle monitor timeout behavior

xavierchanth:
**- What I did**

- mbedtls carries an internal buffer, so if we get a timeout there was truly nothing to read, which means the caller should handle it rather than monitor retrying

**- How I did it**

**- How to verify it**

feat: let the caller handle monitor timeout behavior

### DIFF
--- a/packages/atclient/include/atclient/atclient_utils.h
+++ b/packages/atclient/include/atclient/atclient_utils.h
@@ -30,6 +30,17 @@ int atclient_utils_find_atserver_address(const char *atdirectory_host, const int
  */
 int atclient_utils_populate_atkeys_from_homedir(atclient_atkeys *atkeys, const char *atsign);
 
+/**
+ * @brief find the index of the next character after an atServer / atDirectory's prompt
+ *
+ * @param read_buf buffer to check
+ * @param read_n length of the buffer to check
+ * @param read_i output parameter containing the first token after the prompt
+ *
+ * @return int 0 on success, non-zero on error
+ */
+int atclient_utils_find_index_past_at_prompt(const unsigned char *read_buf, size_t read_n, size_t *read_i);
+
 #ifdef __cplusplus
 }
 #endif
--- a/packages/atclient/src/atclient_utils.c
+++ b/packages/atclient/src/atclient_utils.c
@@ -125,12 +125,16 @@ int atclient_utils_populate_atkeys_from_homedir(atclient_atkeys *atkeys, const c
 exit: { return ret; }
 }
 
-// TODO: unit test
 int atclient_utils_find_index_past_at_prompt(const unsigned char *read_buf, size_t read_n, size_t *read_i) {
   // NOTE: if you change this if, check the second while loop
   // it depends on this guard clause
   *read_i = 0;
   if (read_n != 0 && read_buf[0] != '@') { // Doesn't start with a prompt
+    return 0;
+  }
+
+  if (read_n >= 5 && strncmp((const char *)read_buf, "@null", 5) == 0) {
+    *read_i = 1;
     return 0;
   }
 
--- a/packages/atclient/src/atclient_utils.c
+++ b/packages/atclient/src/atclient_utils.c
@@ -124,3 +124,30 @@ int atclient_utils_populate_atkeys_from_homedir(atclient_atkeys *atkeys, const c
 
 exit: { return ret; }
 }
+
+// TODO: unit test
+int atclient_utils_find_index_past_at_prompt(const unsigned char *read_buf, size_t read_n, size_t *read_i) {
+  // NOTE: if you change this if, check the second while loop
+  // it depends on this guard clause
+  *read_i = 0;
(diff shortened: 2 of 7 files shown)


packages/atclient/src/atclient_utils.c
```diff
@@ -124,3 +124,29 @@ int atclient_utils_populate_atkeys_from_homedir(atclient_atkeys *atkeys, const c
 
 exit: { return ret; }
 }
+
+int atclient_utils_find_index_past_at_prompt(const unsigned char *read_buf, size_t read_n, size_t *read_i) {
```
xavierchanth, line 128: Moved from a private method in atclient.c to public in atclient_utils so that it can be called from monitor.

packages/atclient/src/connection.c
```diff
@@ -435,9 +411,9 @@ int atclient_connection_send(atclient_connection *ctx, const unsigned char *src,
...
     goto exit;
   }
-
+  read_n -= read_i;
   // copy to recv, discarding the prompt
-  memcpy(recv, read_buf + read_i, read_n - read_i);
+  memcpy(recv, read_buf + read_i, read_n);
   free(read_buf);
   recv[read_n - 1] = '\0'; // null terminate the string
   *recv_len = read_n;
```
xavierchanth, line 413: returned buffer size was not being computed correctly.

packages/atclient/src/connection.c
```diff
@@ -602,12 +578,32 @@ int atclient_connection_read(atclient_connection *ctx, unsigned char **value, si
...
+  if (*value == NULL) {
+    free(read_buf);
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to allocate memory for the final output buffer\n");
+    goto exit;
+  }
+  memcpy(*value, read_buf + read_i, read_n);
+  free(read_buf);
+  *value[read_n - 1] = '\0'; // null terminate the string
+  *value_len = read_n;
+
```
xavierchanth, line 600: make sure that we strip prompts in atclient_connect_read, previously this was only done in atclient_connection_send.

Also ensures that the original read buffer is freed in full, so we need a new allocation here. I know how to remove this extra allocation later, but for now, this is the quickest path to robust connections without redesigning how memory management is handled.

packages/atclient/src/monitor.c
```diff
@@ -212,11 +213,21 @@ bool atclient_monitor_is_connected(atclient *monitor_conn) {
...
+  size_t read_i;
+  ret = atclient_utils_find_index_past_at_prompt((unsigned char *)original, original_len, &read_i);
+  if (ret != 0) {
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_DEBUG, "Failed to parse the message: %s\n", original);
+    goto exit;
+  }
+  original = original + read_i;
+  original_len = original_len - read_i;
+  original[original_len - 1] = '\0';
+
```
xavierchanth: Strip prompts from the monitor message, this can happen when the last thing sent or received over the connection was not a notification.

i.e. the first notification after sending monitor or another verb.

packages/atclient/src/socket_mbedtls.c
```diff
@@ -334,15 +333,8 @@ int atclient_tls_socket_read_until_char(struct atclient_tls_socket *socket, unsi
...
-        timeout_count++;
-        if (timeout_count == MAX_READ_TIMEOUTS) {
-          atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_DEBUG, "Failed to read the full message after %d attempts\n",
-                       MAX_READ_TIMEOUTS);
-          return ATCLIENT_SSL_TIMEOUT_EXITCODE;
-        }
-        usleep(1000);
+      case MBEDTLS_ERR_SSL_TIMEOUT:
+        return ATCLIENT_SSL_TIMEOUT_EXITCODE;
         break;
```
xavierchanth, line 338: remove this, let the caller handle it so they have control over the speed of the connection and frequency of retries.

packages/atclient/include/atclient/atclient_utils.h
```diff
@@ -30,6 +30,17 @@ int atclient_utils_find_atserver_address(const char *atdirectory_host, const int
...
+/**
+ * @brief find the index of the next character after an atServer / atDirectory's prompt
+ *
+ * @param read_buf buffer to check
+ * @param read_n length of the buffer to check
+ * @param read_i output parameter containing the first token after the prompt
+ *
+ * @return int 0 on success, non-zero on error
+ */
+int atclient_utils_find_index_past_at_prompt(const unsigned char *read_buf, size_t read_n, size_t *read_i);
```
xavierchanth, line 42: This function has 11 positive test cases as of [8e73faf](https://github.com/atsign-foundation/at_c/pull/479/commits/8e73fafe2001d57c4fb2d8698b0aef18030d76c2)

JeremyTubongbanua (approved):

PR #488, opened 2024-12-30
feat: add test for RSA 4096 operability

TylerTrott:
**- What I did**
Added test that verifies at_c's capability of withstanding a 512-byte key

**- How I did it**
Updated verify func to determine if 256 or 512-byte key is used, created a test that uses a 512-byte generated public key

**- How to verify it**
Run the tests and verify that tests pass with
```
cmake -S . -B build -DATSDK_BUILD_TESTS="unit" && cmake --build build --target all && ./build/packages/atchops/tests/test_rsa_verify
```

feat: tests for RSA 4096


packages/atchops/include/atchops/rsa.h
```diff
@@ -68,16 +69,6 @@ int atchops_rsa_decrypt(const atchops_rsa_key_private_key *private_key, const un
...
-/**
- * @brief generate an RSA keypair
- *
- * @param public_key the public key struct to populate, should be initialized first
- * @param private_key the private key struct to populate, should be initialized first
- * @param keysize the size of the key to generate, e.g. 2048
- */
-int atchops_rsa_generate(atchops_rsa_key_public_key *public_key, atchops_rsa_key_private_key *private_key,
-                         const atchops_md_type md_type);
-
```
xavierchanth, line 80: This was removed from the public API with no replacement for its functionality.

xavierchanth (changes requested):
See my comment

packages/atchops/include/atchops/rsa.h
TylerTrott: It is redundant code, tests pass without it as no code utilises the func. I believe @JeremyTubongbanua has a different way to generate key pair elsewhere

JeremyTubongbanua: https://github.com/atsign-foundation/at_c/blob/9208717af456f1422d3a92282473b6535562ac84/packages/atchops/include/atchops/rsa_key.h#L70

generation code is in rsa_key.h/.c

xavierchanth: Have we added support for specifying the size of the key when we generate?

TylerTrott: Not from an end-developer PoV, I suppose. The way the code is structured allows for 256 or 512-sized keys dynamically and is documented in the code, but if the sdk is utilised in the CLI, I don't think this would be known

xavierchanth: at_c is library only, consumers of at_c may be a CLI but not at_c itself

But, from what I can see in the atchops_rsa_key_generate function, there's no parameter to specify a keysize.

JeremyTubongbanua: Yeah

It seems like the generate function generates a 256 byte modulus (n value) so it only generates a rsa2048 key.

JeremyTubongbanua: Tyler see https://github.com/atsign-foundation/at_c/blob/9208717af456f1422d3a92282473b6535562ac84/packages/atchops/src/rsa_key.c#L244 

Line 244 of rsa_key.c

That should give you a hint as to how to generate a RSA-4096 key

xavierchanth: Ok, since this PR is for tests, we can do that change in another PR if you want

xavierchanth (approved):

packages/atchops/include/atchops/rsa.h
TylerTrott: Oh interesting, I wonder what's gone wrong with my test then

xavierchanth: Well, for generate, it's probably only testing rsa2048. but for hard-code 4096 keys in the tests, everything (in theory) should be working fine

xavierchanth: It would be nice to decrypt the ciphertext after, and check that the value is the same though.

xavierchanth: Or since the key is hardcoded in the tests at least check against the expected ciphertext

TylerTrott: > It would be nice to decrypt the ciphertext after, and check that the value is the same though.

I'll write a test for that such that it'll check against what the expected ciphertext would be

PR #490, opened 2025-01-03
test: fix flaking unit test and make functional tests more adaptable to different atsigns

JeremyTubongbanua:
closes #492 
closes #491 
closes #489 

**- What I did**

Regarding `.github/workflows/tests.yaml`:
- Added `-VV` to unit tests
- Setting the test dir to the directory instead of using working-directory to make it easier to copy and paste it when testing locally
- Added `check_docker_readiness` program that checks if virtual environment has been started and also is PKAM ready.

Regarding `atclient.c`
- Fixed a segfault possibility if err_msg is NULL, then *err_msg is a seg fault.

Regarding `test_atkeys_write.c`
- Moved `test_atkeys_write.c` to functional tests since we test pkam authentication after reading and writing a key file
- Modified the code so that we could use VE atSigns

Regarding `tests/functional_tests/tools`:
- Removed it entirely and we now have `tests/functional_tests/virtualenv`

Regarding `tests/functional_tests`
- Removed `config.c` and moved the helper functions to `helpers.c`
- Moved `functional_tests_set_up_atkeys` to be in `helpers.c`
- Refactored test_atclient_pkam_authenticate to use refactored helper function `functional_tests_set_up_atkeys`

Regarding justfile stuff
- Added parameters to `just.template.env`
- Modified `configure-test-func` in `justfile`

Other changes
- Added `<getopt.h>` to `atactivate_arg_parser.c` cause I was getting an error in Ubuntu 22.04

**- How to verify it**

See this run: https://github.com/atsign-foundation/at_c/actions/runs/12645111162/job/35233754565?pr=490

**- Description for the changelog**
test: fix flaking unit test and make functional tests more adaptable to different atsigns

PR #497, opened 2025-01-07
feat: atserver_message struct

xavierchanth:
**- What I did**

- Added atserver_message struct with protected header file in `src/`
  - Building atclient in Debug mode will include protected symbols
  - Building atclient in Release mode will not include protected symbols (they will still be built internally but not exposed)
    - Note that referencing symbols must be done using the relative include approach
      i.e `#include "atserver_message.h"` NOT `#include <atserver_message.h>` 
- Added parsing and memory management tests for atserver_message struct
- Added flag passthrough to ctest for other test types in just file

**- How I did it**

**- How to verify it**

- Unit tests pass
  - Future changes to integrate this struct into parts of the SDK will yield further validation or necessary changes

feat: atserver_message struct

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Build atSDK
         run: |
-          cmake -S . -B build -DATSDK_BUILD_TESTS="unit"
+          cmake -S . -B build -DATSDK_BUILD_TESTS="unit" -DCMAKE_BUILD_TYPE="Debug"
           cmake --build build --target all
 
       - name: atchops Unit CTest
--- a/packages/atclient/CMakeLists.txt
+++ b/packages/atclient/CMakeLists.txt
@@ -3,6 +3,7 @@ option(ATCLIENT_BUILD_TESTS "Build tests for atclient" OFF)
 
 # Set include directory and file sources
 set(ATCLIENT_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/include)
+set(ATCLIENT_PROTECTED_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/src)
 set(
   ATCLIENT_SOURCES
   ${CMAKE_CURRENT_LIST_DIR}/src/atclient_delete.c
@@ -126,6 +127,13 @@ if(NOT ESP_PLATFORM)
       $<BUILD_INTERFACE:${cjson_SOURCE_DIR}>
       $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   )
+
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    target_include_directories(
+      ${PROJECT_NAME}
+      PUBLIC $<BUILD_INTERFACE:${ATCLIENT_PROTECTED_INCLUDE_DIR}>
+    )
+  endif()
 
   # Link dependencies to library targets
   set(
--- a/packages/atclient/CMakeLists.txt
+++ b/packages/atclient/CMakeLists.txt
@@ -30,6 +30,9 @@ set(
   ${CMAKE_CURRENT_LIST_DIR}/src/notify_params.c
   ${CMAKE_CURRENT_LIST_DIR}/src/request_options.c
   ${CMAKE_CURRENT_LIST_DIR}/src/string_utils.c
+  # protected files
+  ${CMAKE_CURRENT_LIST_DIR}/src/atserver_message.h
+  ${CMAKE_CURRENT_LIST_DIR}/src/atserver_message.c
 )
 
 # Project setup
--- a/packages/atclient/src/atserver_message.c
+++ b/packages/atclient/src/atserver_message.c
@@ -1,0 +1,69 @@
+#include "./atserver_message.h"
+#include <stdlib.h>
+
+const char *atserver_message_get_prompt(struct atserver_message message) {
+  if (message.buffer == NULL || atserver_message_get_prompt_len(message) == 0) {
+    return NULL;
+  }
+  return message.buffer;
+}
+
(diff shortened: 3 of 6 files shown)


packages/atclient/CMakeLists.txt
```diff
@@ -124,6 +128,13 @@ if(NOT ESP_PLATFORM)
       $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   )
 
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    target_include_directories(
+      ${PROJECT_NAME}
+      PUBLIC $<BUILD_INTERFACE:${ATCLIENT_PROTECTED_INCLUDE_DIR}>
+    )
+  endif()
+
```
xavierchanth, line 137: Include the protected dir if we are in debug mode - compile_commands and tests should be generated / built in debug mode.

Customer facing package builds should be generated in release mode

packages/atclient/src/atserver_message.c
```diff
@@ -0,0 +1,71 @@
+#include "./atserver_message.h"
+#include <stdlib.h>
+
+uint16_t atserver_message_get_body_len(struct atserver_message message) {
+  return message.len - message.token_len - message.prompt_len;
+}
```
xavierchanth, line 6: Body length is derived from other metadata

packages/atclient/src/atserver_message.c
```diff
@@ -0,0 +1,71 @@
...
+
+  return message.buffer + message.prompt_len;
+}
+
+char *atserver_message_get_body(struct atserver_message message) {
+  if (message.buffer == NULL || atserver_message_get_body_len(message) == 0) {
+    return NULL;
+  }
+  return (char *)message.buffer + message.token_len + message.prompt_len;
+}
```
xavierchanth, line 28: string getters for various parts of the buffer. The first two are intentionally const, but the third should be allowed to be edited by other parts of the SDK in case we want to do in-place manipulation.

packages/atclient/src/atserver_message.c
```diff
@@ -0,0 +1,71 @@
...
+    token_len++; // corrected for 0 based indexing
+  }
+
+  if (prompt_len > UINT8_MAX || token_len > UINT8_MAX) {
+    // Parse error, by specification they should never come close to exceeding UINT8_MAX
+    return (struct atserver_message){NULL, 0, 0, 0};
+  }
+
+  return (struct atserver_message){buffer, len, prompt_len, token_len};
+}
```
xavierchanth, line 64: Parse function takes a buffer and length, then parses the prompt & token position, returns an atserver_message. returns with values of `{NULL,0,0,0}` if parsing failed

packages/atclient/tests/test_atserver_message.c
```diff
@@ -0,0 +1,245 @@
...
+  }
+  return ret;
+}
+
+#define expect_uint(TEST_TAG, EXP, ACT)                                                                                \
+  if (EXP != ACT) {                                                                                                    \
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "%s: incorrect value (expected: %u, actual: %u)\n", TEST_TAG, EXP, \
+                 ACT);                                                                                                 \
+    return 1;                                                                                                          \
+  }
```
xavierchanth, line 46: macro for expecting two unsigned int types to be equal, otherwise log the error and return 1

packages/atclient/tests/test_atserver_message.c
```diff
@@ -0,0 +1,245 @@
+#include "atlogger/atlogger.h"
+#include "atserver_message.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define BASE_TAG(X) "test_atserver_message " X
```
xavierchanth, line 8: macro to easily generate unique tags for each test function

packages/atclient/tests/test_atserver_message.c
```diff
@@ -0,0 +1,245 @@
...
+  atlogger_set_logging_level(ATLOGGER_LOGGING_LEVEL_INFO);
+
+  ret += test_1a_long_prompt();
+  ret += test_1b_short_prompt();
+  ret += test_1c_no_prompt();
+  ret += test_2a_no_token();
+  ret += test_3a_no_body();
+  ret += test_4a_empty_message();
+  ret += test_5a_heap();
+  ret += test_5b_bad_parse_heap();
```
xavierchanth, line 33: All of these functions return 1 or 0 so ret will not overflow.

xavierchanth:
This will have merge conflicts with #490 , we should merge that PR first.

JeremyTubongbanua:
Instead of having the header files live in `src/` why not have `include/atclient/protected/` ?

xavierchanth:
> protected

Because when someone includes the release build of atclient into their project the symbols will still be available. But this way the symbols aren't available at all.

If they auto-import symbols using LSP and don't notice the header naming, it will still cause breaking changes for them.

JeremyTubongbanua:
> > protected
> 
> Because when someone includes the release build of atclient into their project the symbols will still be available. But this way the symbols aren't available at all.
> 
> If they auto-import symbols using LSP and don't notice the header naming, it will still cause breaking changes for them.

Oh okay

What about `include_protected/` at the same level as `src/` and `include/` ?

.github/workflows/tests.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Build atSDK
         run: |
-          cmake -S . -B build -DATSDK_BUILD_TESTS="unit"
+          cmake -S . -B build -DATSDK_BUILD_TESTS="unit" -DCMAKE_BUILD_TYPE="Debug"
```
xavierchanth, line 20: nvm, thought this would be a conflict but it wasn't changed in #490

JeremyTubongbanua (approved):

PR #499, opened 2025-01-08
fix: memory leaks in tests

xavierchanth:
**- What I did**

This PR removes 114 known memory leaks in at_c's tests:

```
Memory Leak - 114
Potential Memory Leak - 9
```

These remaining 9 potential leaks are in the mbedtls_ssl_handshake, and it doesn't seem to be user-controlled memory, I've tried force freeing different pieces of state and the defects remain. My assumption is that these are false positives. They are subject to further investigation / testing in the future to confirm.

Note that some leaks are in the package code, and some are limited to some needed cleanup in just the tests, both types are addressed.

**- How I did it**

**- How to verify it**

fix: memory leaks in tests

### DIFF
--- a/packages/atclient/src/connection.c
+++ b/packages/atclient/src/connection.c
@@ -747,6 +747,7 @@ static int atclient_connection_set_host(atclient_connection *ctx, const char *ho
     return ret;
   }
 
+  char *old_host = ctx->host;
   /*
    * 2. Allocate memory for the host
    */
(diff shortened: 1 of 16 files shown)


packages/atclient/src/connection.c
```diff
@@ -757,6 +758,7 @@ static int atclient_connection_set_host(atclient_connection *ctx, const char *ho
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to allocate memory for host\n");
     goto exit;
   }
+  free(old_host);
```
xavierchanth, line 761: We only want to free this piece of memory if the allocation for the new host succeeds

JeremyTubongbanua (approved):

PR #510, opened 2025-01-13
feat: standalone atdirectory communication

xavierchanth:
**- What I did**

- Separated atdirectory parsing & connection from connection.c
- unit test for parsing part
- call atdirectory from atclient_utils_find_atserver_address

**- How I did it**

**- How to verify it**

feat: standalone atdirectory communication


packages/atclient/src/atclient.c
```diff
@@ -238,10 +238,6 @@ int atclient_pkam_authenticate(atclient *ctx, const char *atsign, const atclient
 
   bool should_free_atserver_host = false;
 
-  // expected result on a successful login
-  size_t expected_len = 1 + strlen(atsign) + strlen("@data:success");
-  char expected_buf[expected_len];
-
```
xavierchanth, line 244: this was dead code

cpswan (approved):

packages/atclient/src/atdirectory.c
```diff
@@ -0,0 +1,152 @@
+#include "atdirectory.h"
+#include "atlogger/atlogger.h"
+#include <string.h>
+#define TAG "atdirectory"
+
+struct atdirectory_connection atdirectory_connection_create(const char *host, const uint16_t port) {
```
JeremyTubongbanua, line 6: Why are we going with `create` now instead of the `init` pattern?

xavierchanth: > Why are we going with `create` now instead of the `init` pattern?

init is commonly used to initialize a structure which has already been allocated.

create returns a struct which may allocate memory and populate values, or it can fail, in which case those values in the struct are null.

JeremyTubongbanua: So why are we forcing the developer to dynamically allocate `atdirectory_connection` ?

xavierchanth: We aren’t.

It’s private, so the developer will not have access to it.

The struct is statically allocated, only the host string is dynamically allocated at initialization

PR #511, opened 2025-01-13
feat: CMakePresets & CPack Workflow

xavierchanth:
**- What I did**

- Added CMakePresets file with presets for ci & static-release configuration
  - usage: `cmake --preset <preset-name>`
- Added build & package presets for static releases (zip & tar.gz)
- Added workflow for a static release which combines all of the above
  - usage: `cmake --workflow --preset package-static`
  - This runs the configure & build steps, then uses the `package-source` dir to generate an appropriate source file structure and then generates zip & tar.gz files.

> For some reason, the cJSON build always tries to install system headers, so I've set the install directory to `package-source/install` to prevent cmake from trying to install these to a privileged directory.

**- How I did it**

**- How to verify it**

Checkout & execute this command:
`cmake --workflow --preset package-static`

feat: CMakePresets & CPack Workflow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ cmake_install.cmake
 compile_commands.json
 
 build/
+package-source/
 install/
 bin/
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,90 @@
+{
+  "version": 8,
+  "configurePresets": [
+    {
+      "name": "ci-base",
+      "hidden": true,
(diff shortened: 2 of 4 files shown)


xavierchanth:
@cpswan next step is to integrate this with the existing release flow, (and the CI configuration with the CI flow)

Before I do that, can you generate a source tar locally, and see if the generated tar contains everything you need to continue with Yocto & openwrt?

cpswan (approved):

PR #513, opened 2025-01-14
feat: cpack with source

xavierchanth:
**- What I did**

- Removed CPackSourceConfig.cmake in favor of cmake/CPackSetup.cmake
  - This script contains all of the CPack config and gets included in the CMakeLists.txt
  - This will generate 2 additional cmake files for packaging in the build directory
  - This also generates 2 addition targets (package and package_source)
- Updated CMakePresets.json to have 2 workflows:
  - `cmake --workflow --preset package-static`
  - `cmake --workflow --preset package-source`

**- How I did it**

**- How to verify it**

feat: cpack with source


cmake/CPackSetup.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+# source cpack configuration
+set(CPACK_SOURCE_PACKAGE_FILE_NAME atsdk-source)
+set(CPACK_SOURCE_GENERATOR TGZ ZIP)
+
+set(CPACK_SOURCE_INSTALLED_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR};/")
+
+# Yes, you need 4 * '\' to escape literal '.'
+set(
+  CPACK_SOURCE_IGNORE_FILES
```
xavierchanth, line 23: @cpswan please review the list of ignored files. I think I've chosen a decent middle ground of removing all project development specific things, but kept all things that might be relevant to a customer (e.g. examples)

xavierchanth: dir structure: https://gist.github.com/XavierChanth/62310db3f48721c7807309f28ab323b1

xavierchanth:
Another (potential) problem has occurred:
- mbedtls and cjson are not included as source, is this necessary?

xavierchanth:
TODOs (for myself):

Include cjson & mbedtls source, add a configuration option to the CMake which tells the build system to pull them from a path.

packages/atchops/tests/CMakeLists.txt
```diff
@@ -20,3 +20,4 @@ foreach(file ${files})
   target_link_libraries(${filename} PRIVATE atchops atlogger)
   add_test(NAME ${filename} COMMAND $<TARGET_FILE:${filename}>)
 endforeach()
+enable_testing()
```
xavierchanth, line 23: I snuck in some adjustments to enable_testing, so we don't build and run mbedtls & cjson tests all the time.

cmake/find_cjson.cmake
```diff
@@ -1,21 +1,28 @@
...
-    URL_HASH
-      SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
-    # FIND_PACKAGE_ARGS 1.7.17 QUIET CONFIG
-  )
-
+  if(ATSDK_CJSON_PATH)
+    FetchContent_Declare(
+      cjson
+      SOURCE_DIR
+      ${CMAKE_SOURCE_DIR}/${ATSDK_CJSON_PATH}
```
xavierchanth, line 15: When ATSDK_CJSON_PATH is passed (and ATSDK_MBEDTLS_PATH for mbedtls) then cmake uses the relative path to the root CMakeLists.txt's directory.

cmake/CPackSetup.cmake
```diff
@@ -0,0 +1,60 @@
...
+  "\\\\.gitignore"
+  "compile_commands.json"
+)
+
+set(CPACK_INSTALL_SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/CPackSourceDeps.cmake")
+configure_file(
+  "${CMAKE_SOURCE_DIR}/cmake/CPackSourceDeps.cmake"
+  "CPackSourceDeps.cmake"
+  @ONLY
+)
```
xavierchanth, line 53: Added these lines to run an additional script which copies over mbedtls and cjson src...

but it also does so for the static build, I _could_ spend time to figure out how to conditionally do this, but there is virtually no documentation for cpack variables... even some of the ones being used here aren't documented... so I'm not going to spend time on something we don't care about right now.

As for why we can't do this with the exclude list:
Unfortunately there is no way in cmake to specify inverted regular expressions, and generating a recursive globular filter is awful. I already tried, but we need to go multiple levels deep, and performance was already terrible.

xavierchanth:
@cpswan usage:

generate the packaged zips/tar.gz:
```
cmake --workflow --preset package-static
cmake --workflow --preset package-source
# OR
just package-static
just package-source
```

output follows the pattern `./build/release-static/atsdk-{source,static}.{zip,tar.gz}`

To build from the atsdk-source

```
cmake -S . -B <build-dir> \
  -DATSDK_CJSON_PATH=deps/cjson-src \
  -DATSDK_MBEDTLS=deps/mbedtls-src
  # rest of your options...
```

I could also make it so we don't need to pass these arguments, but once again timeboxing it.

cpswan:
`cmake --workflow --preset package-static` is failing for me:

```log
/home/chris/git/github.com/atsign-foundation/at_c/packages/atclient/src/atclient_utils.c: In function ‘atclient_utils_find_atserver_address’:
/home/chris/git/github.com/atsign-foundation/at_c/packages/atclient/src/atclient_utils.c:16:94: error: ‘u_int16_t’ undeclared (first use in this function); did you mean ‘uint16_t’?
   16 |   return atdirectory_lookup_once(atdirectory_host, atdirectory_port, atsign, atserver_host, (u_int16_t *)atserver_port);
      |                                                                                              ^~~~~~~~~
      |                                                                                              uint16_t
/home/chris/git/github.com/atsign-foundation/at_c/packages/atclient/src/atclient_utils.c:16:94: note: each undeclared identifier is reported only once for each function it appears in
/home/chris/git/github.com/atsign-foundation/at_c/packages/atclient/src/atclient_utils.c:16:105: error: expected expression before ‘)’ token
   16 |   return atdirectory_lookup_once(atdirectory_host, atdirectory_port, atsign, atserver_host, (u_int16_t *)atserver_port);
      |                                                                                                         ^
gmake[2]: *** [packages/atclient/CMakeFiles/atclient.dir/build.make:188: packages/atclient/CMakeFiles/atclient.dir/src/atclient_utils.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:702: packages/atclient/CMakeFiles/atclient.dir/all] Error 2
gmake: *** [Makefile:156: all] Error 2
```

cpswan:
`cmake --workflow --preset package-source` seems to work :)

But... I get a few warnings when I configure:

```log
CMake Warning (dev) at /usr/share/cmake-3.28/Modules/GNUInstallDirs.cmake:243 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:33 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

```log
-- [cJSON] fetching package...
CMake Deprecation Warning at deps/cjson-src/CMakeLists.txt:2 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

```log
-- [MbedTLS] fetching package...
CMake Warning (dev) at build/_deps/mbedtls-src/CMakeLists.txt:65 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'ENABLE_PROGRAMS'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

```log
-- Generating done (0.2s)
CMake Warning:
  Manually-specified variables were not used by the project:

    ATSDK_MBEDTLS
```

Also (a nit) - we usually use whatever.`tgz` rather than whatever,`tar.gz`

If we can now get this working for C sshnpd we should be in a much better place for OpenWrt and Yocto :)

xavierchanth:
> Also (a nit) - we usually use whatever.`tgz` rather than whatever,`tar.gz`

We will have to rename externally afterwards, cpack doesn't provide the ability to change the file extension.

For the warnings:

1. looking at the cmake, everywhere we specify GNUInstallDirs, a language is set first, this must be in one of the dependencies
2. This is in cJSON, their project is relying on an older version of cmake, I suspect when (if) the deprecated feature is removed, the community will voice the need to update the cmake policies.
3. Added a policy default for CMP0077 to use the old behaviour in find_mbedtls.cmake.

cpswan:
Can we add these to the build tests (so we catch things like the u_uint thing sooner):

```sh
cmake --workflow --preset package-static
cmake --workflow --preset package-source
```

Then I think we're probably good to go with this PR.

xavierchanth:
> Can we add these to the build tests (so we catch things like the u_uint thing sooner):
> 
> ```shell
> cmake --workflow --preset package-static
> cmake --workflow --preset package-source
> ```
> 
> Then I think we're probably good to go with this PR.

Added here: [ed3f10a](https://github.com/atsign-foundation/at_c/pull/513/commits/ed3f10a8eb04e2b1ecb01f767b8d1ddf23dbe434)

cpswan (dismissed):

xavierchanth:
> Can we add these to the build tests (so we catch things like the u_uint thing sooner):

We may or may not be able to catch this particular one, u_intX_t is non-standardized, uintX_t is. Some compilers / systems will have it, some won't, it's more unpredictable than some of the other issues we've been facing.

xavierchanth:
@cpswan I need another review, there was a typo in the workflow.

One thing I dislike about the new merge experience is that when a workflow fails due to a syntax error, it just doesn't show up at all in the view.

xavierchanth:
> @cpswan I need another review, there was a typo in the workflow.

nvm, looks like ci is reproducing the issues you uncovered.

I'm going to open a PR which is the merge of this branch and the branch I'm using to fix these issues. We can merge them individually, but I'd like to confirm resolution of the issues first

cpswan (approved):

PR #514, opened 2025-01-14
fix: remove private linux header

xavierchanth:
**- What I did**

- Use `limits.h` instead of private linux header `linux/limits.h` on linux

**- How I did it**

**- How to verify it**

fix: remove private linux header

### DIFF
--- a/packages/atauth/src/atauth_fetch_home_dir.c
+++ b/packages/atauth/src/atauth_fetch_home_dir.c
@@ -11,14 +11,6 @@
 #define PATH_SEPARATOR '\\'
 #define PATH_MAX 260 // Max path length for Windows (adjustable if needed)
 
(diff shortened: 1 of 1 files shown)


packages/atauth/src/atauth_fetch_home_dir.c
```diff
@@ -11,14 +11,6 @@
...
 
-// Imports for Linux
-#elif defined(__linux__)
-#include <linux/limits.h>
-#include <pwd.h>
-#include <unistd.h>
-#define PATH_SEPARATOR '/'
-
-// Imports for other platforms
 #else
```
xavierchanth, line 14: macos & linux headers should be the same now

cpswan (approved):

PR #509, opened 2025-01-13
fix: Enable all unit tests

xavierchanth:
**- What I did**

- Enabled all unit tests in github actions instead of just atchops/atclient
- commits before [95d0c5e](https://github.com/atsign-foundation/at_c/pull/509/commits/95d0c5e30aec2262cabf0fa11399b3887e5246c0) are also contained in https://github.com/atsign-foundation/at_c/pull/508
- remove unit tests with segfaults
  - I will be revising the code & write new tests in another PR

**- How I did it**

**- How to verify it**

- Unit tests pass

**- Description for the changelog**
fix: Enable all unit tests

PR #521, opened 2025-01-20
chore: move atauth code in atcommons into atauth

xavierchanth:
**- What I did**

- Move atauth specific code that's currently in atcommons into atauth
- The purpose of this change is primarily so we can drop the cjson dependency from atcommons so it can be a pure C package for future changes
- It also means that tests for atauth specific behavior are actually in atauth
- Moved json.h into atclient, atclient is now the most upstream dependency which contains cjson.
- atcommons is an empty package for now

**- How I did it**

**- How to verify it**

- All tests pass except for ` 25 - test_enroll_namespace_utils (Bus error)` which has a known segfault
  - edit: this test isn't enabled until https://github.com/atsign-foundation/at_c/pull/509 is merged (my next goal for today)

chore: move atauth code in atcommons into atauth


packages/atcommons/include/atcommons/temp.h
xavierchanth, line 1: These placeholders are necessary so atcommons will build despite being an empty package

cpswan (approved):

PR #545, opened 2025-02-03
feat(atcommons): memory management util

### LINKED ISSUE #543: feat(atcommons): memory management utilities
A memory management utilities to help us simplify memory management in some of the difficult areas (some of which do have memory leaks):

### Requirements

- allocate a base object which contains the full list of dynamically allocated memory for a given scope.
- append to the base object with newly added pieces of memory, passing back a handle (index) or -1 for failure
  - ability to specify some options:
    - free always vs only free on function failure (i.e. heap allocated output should only be freed on failure)
    - custom override function for when we have pointers to structs which may have custom free functions
- free_success / free_failure functions (naming to be improved)
- free_index functions to free a single item based on the handle from append

### Implementation details

Main ideas:
- use a predefined stack array (AoS) since the maximum number of pointers is usually known
   and doesn't ever require reallocation.
- use SoA for packing efficiency, since we are using a by index / by handle approach
   this approach makes alot of sense since we are storing a single bool and reduces memory per node by N-1 bits
   on an N bit architecture
- use a (doubly) linked list (ideally the nodes are treated & freed like a stack - freed back to front)
   this allows for each node to be allocated independently, ideal when the upper limit is not known.

### Additional notes:

One of the other main blockers with our use of goto statements to manage cleanup is it limits when we can allocate dynamic stack arrays. They must be done before any gotos since the positions of memory in the stack may potentially be disrupted by them. In our case, they wouldn't be because we have specific rules about how goto statements are and aren't allowed to be used for the same reasons we can't implement dynamic stack arrays after a goto. But the compiler isn't aware enough of how gotos & dynamic stack arrays interact to be able to compile this code safely, even though in theory, it is possible. 

While this doesn't seem like a big deal, this is a problem in some functions where the output size is determined by some function which has potential to fail (and thus is gatekeeped behind a goto). This means we typically reach for heap allocations when we should be reaching for dynamic stack arrays.

xavierchanth:
**- What I did**

closes #543 

Created memory management util in atcommons to help address some leaks in atclient.

- Manages a list of up to 32 objects (the struct itself is stack allocated).
- Supports the case where something should remain allocated on success, but not failure.
- Supports custom free functions via a function pointer otherwise calls `free()`

**- How I did it**

**- How to verify it**

feat(atcommons): memory management util

### DIFF
--- a/packages/atcommons/src/memory_util.c
+++ b/packages/atcommons/src/memory_util.c
@@ -1,0 +1,106 @@
+#include "atlogger/atlogger.h"
+#include <atcommons/memory_util.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define TAG "atcommons_memlist"
+struct atcommons_memlist atcommons_memlist_create(uint32_t len) {
+  // the only way to check this at compile time is the road to macro hell
+  // while it would make more robust code when done right, it would be a waste
+  // of time trying to get it right (at least right now)
+  if (len > ATCOMMONS_MEMLIST_SIZE) {
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR,
+                 "requested length exceeds the maximum allowed: %d; allocating maximum of %d instead\n", len,
+                 ATCOMMONS_MEMLIST_SIZE);
+    len = ATCOMMONS_MEMLIST_SIZE;
+  }
+  return (struct atcommons_memlist){
+      .len = len,
+      .free_on_success = 0,
+      .allocated = 0,
+      .memory = {(struct atcommons_memory_deallocator){.memory = NULL, .free_fn = NULL}},
+  };
+}
+
+int atcommons_memlist_add(struct atcommons_memlist *memlist, void *memory, bool free_on_success, void *free_fn) {
+  if (memory == NULL && free_fn == NULL) {
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "One of memory and free_fn must be non-NULL\n");
+    return 1;
+  }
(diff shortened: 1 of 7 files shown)


cpswan (approved):

packages/atcommons/src/memory_util.c
```diff
@@ -0,0 +1,106 @@
...
+  dealloc->free_fn = NULL;
+  dealloc->memory = NULL;
+  memlist->allocated &= ~(0b1 << index); // set bit at index to 0
+  return 0;
+}
+
+void atcommons_memlist_failure_free(struct atcommons_memlist *memlist) {
+  for (uint32_t i = 0; i < memlist->len; i++) {
+    if ((memlist->allocated & (0b1 << i)) != 0) {
+      atcommons_memlist_index_free(memlist, i);
```
xavierchanth, line 93: ignored ret value: free by index can only fail due to invalid input. Which can only happen from within this function if the struct was modified directly (which shouldn't ever be done)

packages/atcommons/src/memory_util.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+}
+
+void atcommons_memlist_success_free(struct atcommons_memlist *memlist) {
+  // create a mask of bits where memory is allocated and free_on_success is true
+  ATCOMMONS_MEMLIST_BITMASK_TYPE bitmask = memlist->allocated & memlist->free_on_success;
+  for (uint32_t i = 0; i < memlist->len; i++) {
+    if ((bitmask & (0b1 << i)) != 0) {
+      atcommons_memlist_index_free(memlist, i);
```
xavierchanth, line 103: same as above comment

PR #553, opened 2025-02-07
fix: replace newline with null terminator when parsing atserver_message

xavierchanth:
**- What I did**

- Ensure atserver_message's buffer is null-terminated
- Disabled the one suppression we had added to the test suite

**- How I did it**

- During parsing of atserver_message, replace the newline with a null-terminator
  - Chosen not to do this on the socket level as replacing a character
    would alter socket data and the caller would have to realloc and append
    the character if they want it to be included

Note: it's okay for atserver_message to modify the buffer, because it is meant to take ownership of the memory passed into it. atserver_message is meant to store a buffer and the positions of the prompt, token (data,error,notification) and the message body.

**- How to verify it**

- Ensure all tests are passing

fix: replace newline with null terminator when parsing atserver_message

### DIFF
--- a/packages/atclient/src/atnotification.c
+++ b/packages/atclient/src/atnotification.c
@@ -14,7 +14,26 @@ static int atclient_atnotification_from_cjson_node(atclient_atnotification *noti
 #endif
 
 void atclient_atnotification_init(atclient_atnotification *notification) {
-  memset(notification, 0, sizeof(atclient_atnotification));
+  // most of the fields are char* so manually set each to NULL
+  // in some niche cases, NULL != 0
+  notification->id = NULL;
+  notification->from = NULL;
+  notification->to = NULL;
+  notification->key = NULL;
+  notification->value = NULL;
+  notification->operation = NULL;
+  notification->message_type = NULL;
+  notification->enc_key_name = NULL;
+  notification->enc_algo = NULL;
+  notification->iv_nonce = NULL;
+  notification->ske_enc_key_name = NULL;
+  notification->ske_enc_algo = NULL;
+  notification->decrypted_value = NULL;
+  // Since there are very few non char* values, set each one manually
+  notification->epoch_millis = 0;
+  notification->is_encrypted = false;
+  notification->_initialized_fields[0] = 0;
+  notification->_initialized_fields[1] = 0;
 }
 
 void atclient_atnotification_free(atclient_atnotification *notification) {
--- a/packages/atclient/src/atserver_message.c
+++ b/packages/atclient/src/atserver_message.c
@@ -1,5 +1,9 @@
 #include "./atserver_message.h"
+#include "atlogger/atlogger.h"
 #include <stdlib.h>
+#include <string.h>
+
+#define TAG "atserver_message"
 
 size_t atserver_message_get_body_len(struct atserver_message message) {
   return message.len - message.token_len - message.prompt_len;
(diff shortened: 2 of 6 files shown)


packages/atclient/src/atnotification.c
```diff
@@ -14,7 +14,26 @@ static int atclient_atnotification_from_cjson_node(atclient_atnotification *noti
...
+  notification->enc_algo = NULL;
+  notification->iv_nonce = NULL;
+  notification->ske_enc_key_name = NULL;
+  notification->ske_enc_algo = NULL;
+  notification->decrypted_value = NULL;
+  // Since there are very few non char* values, set each one manually
+  notification->epoch_millis = 0;
+  notification->is_encrypted = false;
+  notification->_initialized_fields[0] = 0;
+  notification->_initialized_fields[1] = 0;
```
xavierchanth, line 36: Found this during investigation, while on most systems `NULL == 0`, this isn't always the case. As well, if I recall correctly, some of the MS tooling changes the value of NULL to detect for NULL vs 0 related issues.

xavierchanth:
Setting back to draft, some more edge cases in the SDK to fix

packages/atclient/src/atserver_message.c
```diff
@@ -29,7 +33,19 @@ char *atserver_message_get_body(struct atserver_message message) {
...
+  }
+
+  if (len > 2 && buffer[len - 2] == '\r' && buffer[len - 1] == '\n') {
+    buffer[len - 2] = 0;
+    len -= 2;
+  } else if (len > 1 && buffer[len - 1] == '\n') {
+    buffer[len - 1] = 0;
+    len--;
+  } else {
+    atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Invalid message: does not end with a newline\n");
```
xavierchanth, line 48: replace the `\r\n` or `\n` with a null terminator, then adjust the string length.

If there is no `\r\n` or `\n` then we reject the message as a failed parse. If parsed directly from the socket layer, assuming the server message is not malformed due to some weird socket / buffer alignment issue (which I've already spent some time on to avoid it happening), the parse should always parse.

packages/atclient/tests/test_atserver_message.c
```diff
@@ -47,8 +47,10 @@ int main() {
...
 uint8_t test_1a_long_prompt() {
   const char *TAG = BASE_TAG("1a");
-  char *buffer = "@foobar@data:baz";
-  const uint16_t len = strlen(buffer);
+
+  const uint16_t len = 17;
+  char buffer[len];
+  memcpy(buffer, "@foobar@data:baz\n", len);
 
   struct atserver_message message = atserver_message_parse(buffer, len);
```
xavierchanth, line 55: Made all of these buffers in the tests non-constant now that atserver_message_parse modifies them.

packages/atclient/tests/test_atserver_message.c
```diff
@@ -233,7 +238,7 @@ uint8_t test_5b_bad_parse_heap() {
   expect_uint("len", 0, message.len);
 
   // ensure original heap buffer is intact
-  if (strncmp(heap_buffer, static_buffer, len) != 0) {
+  if (strncmp(heap_buffer, static_buffer, len - 1) != 0) {
```
xavierchanth, line 241: -1 to avoid comparing the `\n` which now should be replaced with a null-terminator

xavierchanth:
@cpswan ready for review

cpswan (approved):

PR #577, opened 2025-04-01
feat(notifications): set pointer of decrypted_value to value when is_encrypted is false.

### LINKED ISSUE #559: Make atnotification->decrypted_value point to atnotification->value when it is a non-encrypted notification
The way we currently handle non-encrypted notifications in monitor is that monitor_read populates `atnotification->value` to store the value, since no decryption is needed. Therefore, `atnotification->decrypted_value` remains null.

The proposed change is to have `atnotification->decrypted_value` point to `atnotification->value`, since it is non-encrypted anyways. This is a developer experience change.

xavierchanth:
closes #559 

### Some Context

notification contains:

```c
bool is_encrypted;
char* value; // the buffer of value directly from the notification payload
char* decrypted_value; // the buffer of the decrypted value
// ... more irrelevant things
```

**- What I did**

- Previously, decrypted value is NULL if the notification was not encrypted. This change makes decrypted_value point to the same buffer as value when we parse notifications.
- Upgraded cJSON to 1.7.18, to deal with cmake deprecations.

**- How I did it**

- Made it so that decrypt_notification is called for non-encrypted notifications
- Only free decrypted_value in a notification if is_encrypted is true, other wise we get a double free since value and decrypted_value point to the same memory.
- Added code to point decrypted_value at value when is_encrypted is false.
- During cleanup ensure that decrypted_value is handled before is_encrypted, since there is a dependency on it (comment added to code to avoid regressions)

**- How to verify it**

Tests should pass.

There was already a test for non-encrypted notifications, so I extended it to check both value and decrypted_value for the expected value.

feat(notifications): set pointer of decrypted_value to value when is_encrypted is false.

### DIFF
--- a/cmake/find_cjson.cmake
+++ b/cmake/find_cjson.cmake
@@ -17,9 +17,10 @@ if(NOT TARGET cjson)
   else()
     FetchContent_Declare(
       cjson
-      URL https://github.com/DaveGamble/cJSON/archive/refs/tags/v1.7.17.zip
+      URL https://github.com/DaveGamble/cJSON/archive/refs/tags/v1.7.18.zip
       URL_HASH
-        SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
+        # SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
+        SHA256=cc6d93cc3b659037c34193ecc7be5a874a18c2ac67b24efe82db6a759b486b5d # hash for v1.7.18 .zip release source code
       # FIND_PACKAGE_ARGS 1.7.17 QUIET CONFIG
     )
   endif()
--- a/.github/workflows/build_source_tests.yaml
+++ b/.github/workflows/build_source_tests.yaml
@@ -23,8 +23,12 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Show compiler version
-        run: ${{ matrix.cc }} --version
+      - name: Show tooling versions
+        run: |
+          echo ${{ matrix.cc }} version
+          ${{ matrix.cc }} --version
+          echo cmake version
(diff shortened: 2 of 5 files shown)


xavierchanth:
New cmake version has caused the build to fail, the cjson build code is now deprecated, working on a patch.

xavierchanth:
Alright, let's see if Copilot is useful at reviewing this code. It was awful when it came to writing C code.

xavierchanth:
> ## Pull Request Overview
> 
> This PR updates the notifications module by adjusting the behavior for non-encrypted notifications so that decrypted_value points to the value buffer, avoiding double free errors, and improves test precision. Additionally, the build workflow has been updated to display both the compiler and cmake versions for improved tooling debugging.
> Files not reviewed (4)
> 
>     * **cmake/find_cjson.cmake**: Language not supported
> 
>     * **packages/atclient/src/atnotification.c**: Language not supported
> 
>     * **packages/atclient/src/monitor.c**: Language not supported
> 
>     * **tests/functional_tests/tests/test_atclient_monitor.c**: Language not supported

Good job copilot /s

cmake/find_cjson.cmake
```diff
@@ -17,9 +17,10 @@ if(NOT TARGET cjson)
   else()
     FetchContent_Declare(
       cjson
-      URL https://github.com/DaveGamble/cJSON/archive/refs/tags/v1.7.17.zip
+      URL https://github.com/DaveGamble/cJSON/archive/refs/tags/v1.7.18.zip
       URL_HASH
-        SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
+        # SHA256=51f3b07aece8d1786e74b951fd92556506586cb36670741b6bfb79bf5d484216 # hash for v1.7.17 .zip release source code
+        SHA256=cc6d93cc3b659037c34193ecc7be5a874a18c2ac67b24efe82db6a759b486b5d # hash for v1.7.18 .zip release source code
```
xavierchanth, line 23: 1.7.17 fails to build with latest cmake in the GHA runner, as cmake <3.5 has been deprecated.

gkc (approved):
lgtm

PR #585, opened 2025-05-09
fix: openwrt upstream build

xavierchanth:
**- What I did**

- Added additional warning checks:
  - `-Wtype-limits`
  - `-Wimplicit-fallthrough`
  - `-Wmissing-field-initializers`
- Moved warning compiler options from justfile / CMakePresets to CMakeLists.txt / .clangd
- Fixed any newly revealed warnings (they align with what we saw in openwrt build)

**- How I did it**

**- How to verify it**

fix: openwrt upstream build

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -2,7 +2,17 @@ CompileFlags:
   # Use gcc in development for better diagnostics
   # & Ensure that compilation doesn't fail due to warnings
   Compiler: gcc
-  Add: [-Wno-error, -xc, -std=c99, -DATSDK_DEBUG_MODE]
+  Add:
+    - -xc
+    - -std=c99
+    - -DATSDK_DEBUG_MODE
+    - -Wno-error
+    - -Wall
+    - -Wextra
+    - -Werror-implicit-function-declaration
+    - -Wtype-limits
+    - -Wimplicit-fallthrough
+    - -Wmissing-field-initializers
 
 Index:
   # Enable background indexing for better symbol information
--- a/justfile
+++ b/justfile
@@ -16,12 +16,12 @@ unit_dir := quote(justfile_directory() / "build/unit-") + postfix
 func_dir := quote(justfile_directory() / "build/func-") + postfix
 memcheck_dir := quote(justfile_directory() / "build/memcheck-") + postfix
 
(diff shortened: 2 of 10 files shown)


packages/atclient/src/monitor.c
```diff
@@ -39,6 +38,7 @@ void atclient_monitor_message_free(atclient_monitor_message *message) {
   case ATCLIENT_MONITOR_ERROR_DECRYPT_NOTIFICATION:
     atclient_atnotification_free(message->notification);
     free(message->notification);
+    __attribute__((fallthrough));
```
xavierchanth, line 41: Tells the compiler that this is an explicit fallthrough not an implicit one where a `break` is missing.

packages/atclient/src/connection.c
```diff
@@ -841,12 +841,6 @@ static int atclient_connection_set_port(atclient_connection *ctx, const uint16_t
     return ret;
   }
 
-  if (port < 0) {
```
xavierchanth, line 844: port was changed to an unsigned 16 bit integer, so it can never be less than zero.

justfile
```diff
@@ -16,12 +16,12 @@ unit_dir := quote(justfile_directory() / "build/unit-") + postfix
 func_dir := quote(justfile_directory() / "build/func-") + postfix
 memcheck_dir := quote(justfile_directory() / "build/memcheck-") + postfix
 
-debug_c_flags := "-std=c99 -Wall -Wextra -Werror-implicit-function-declaration -DATSDK_DEBUG_MODE=1"
```
xavierchanth, line 19: Moved these from the just file into the cmake config

.clangd
```diff
@@ -2,7 +2,17 @@ CompileFlags:
   # Use gcc in development for better diagnostics
   # & Ensure that compilation doesn't fail due to warnings
   Compiler: gcc
-  Add: [-Wno-error, -xc, -std=c99, -DATSDK_DEBUG_MODE]
+  Add:
```
xavierchanth, line 5: Added more checks to clangd lsp analysis.

cpswan (approved):

PR #590, opened 2025-05-27
fix: Don't search for everest

cpswan:
OpenWrt doesn't make use of everest.h (formally verified version of ecc implementation that can be used in mbedtls)

We also had flags in place that were preventing libraries from being found in a cross compiler environment

**- What I did**

* Removed dependency on `everest/everest.h`
* Removed CMake flags that narrow search

**- How to verify it**

I've build a source package against this branch (and forthcoming PR to noports) and then built an OpenWrt package from it.

fix: Don't search for everest


xavierchanth (dismissed):
@cpswan do you want to bump the version #

cpswan:
> @cpswan do you want to bump the version #

Yes, thanks @XavierChanth fixed in https://github.com/atsign-foundation/at_c/pull/590/commits/f9b682b81dcd987e04ab8803931c9b60b7578d8a

xavierchanth (approved):

PR #592, opened 2025-06-02
fix: apply patch to cjson CMakeLists.txt so it will build with cmake 4.0

xavierchanth:
**- What I did**

Applied the following patch to cjson's CMakeLists.txt when we fetch it:

```patch
--- i/CMakeLists.txt
+++ w/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(CMAKE_LEGACY_CYGWIN_WIN32 0)
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.24)
```

cmake <3.5 is deprecated by cmake 4.0, so we need this version number to be at least `3.5`, I set it to `3.24` because that is our standard base line for cmake version.

I did not bump versions anywhere as we don't need to publish this until we have the next functional change.

**- How I did it**

**- How to verify it**

fix: apply patch to cjson CMakeLists.txt so it will build with cmake 4.0


cmake/cjson.patch
```diff
@@ -0,0 +1,6 @@
+--- i/CMakeLists.txt
++++ w/CMakeLists.txt
+@@ -1,2 +1,2 @@
```
cpswan, line 5:
```suggestion
# cmake <3.5 is deprecated by cmake 4.0,
# so we need this version number to be at
# least 3.5. Set to 3.24 as that is our standard
# base line for cmake version.
@@ -1,2 +1,2 @@
```

cpswan (changes requested):
LGTM. Though I think we should leave evidence in the patch (not just this PR) for why it's there.

xavierchanth:
> LGTM. Though I think we should leave evidence in the patch (not just this PR) for why it's there.

I had tried to use sed yesterday instead of a patch, targeting the exact line to prevent any unintentional replacements. It seems that wrapping the command in bash -c to get a shell works well.

I have documented this with comments in the cmake file.

The main reason for trying this method is to avoid the need to install the patch command in our C-buildimage docker files.

xavierchanth:
Darn, I missed this comment: https://github.com/atsign-foundation/noports/pull/1960#issuecomment-2934311759

I'm going to backout the [b134eaa](https://github.com/atsign-foundation/at_c/pull/592/commits/b134eaa3e1092a41f107d3abcb12d8bd6eca7b17) changes

cpswan (approved):

PR #593, opened 2025-06-03
fix: cjson patch with sed

xavierchanth:
**- What I did**

Replaced patch command with sed script to patch cjson.

Both patch and sed are different in GNU & Unix, but I am already familiar with the differences in sed, so I switched to that.

From reading the patch man pages, I couldn't find anything which indicated that the patch in cmake should fail on MacOS but it is.

The sed command will only work from within a shell, this should work on any platform which meets the following three things:

1. /bin/sh is a posix compliant shell AND
2. sed is GNU sed OR (linux)
3. sed is Unix sed AND uname = 'Darwin' (macos)

**- How I did it**

**- How to verify it**

Tested on:

- Macos 15.5, with CMake 4.0.1 & gcc (Apple clang version 16.0.0)
- Arch latest, with CMake 4.0.2 & clang 19.1.7

fix: cjson patch with sed

### DIFF
--- a/cmake/cjson-patch.sh
+++ b/cmake/cjson-patch.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# This PATCH_COMMAND script performs the following:
+# Replaces the following line in the cJSON CMakeLists.txt:
+# `cmake_minimum_required(VERSION 3.0)`
+# With the line:
+# `cmake_minimum_required(VERSION 3.24)`
+# This allows us to build it with CMake v4
+if [ "$(uname)" = 'Darwin' ]; then
+  # Unix sed
+  sed -e '2s/0/24/' -i '' CMakeLists.txt
+else
+  # GNU sed
+  sed -e '2s/0/24/' -i CMakeLists.txt
+fi
--- a/cmake/cjson.patch
+++ b/cmake/cjson.patch
@@ -1,6 +1,0 @@
---- i/CMakeLists.txt
-+++ w/CMakeLists.txt
-@@ -1,2 +1,2 @@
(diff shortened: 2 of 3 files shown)


xavierchanth:
Going to test on macos before setting as ready to review

cmake/cjson-patch.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# This PATCH_COMMAND script performs the following:
+# Replaces the following line in the cJSON CMakeLists.txt:
+# `cmake_minimum_required(VERSION 3.0)`
+# With the line:
+# `cmake_minimum_required(VERSION 3.24)`
+# This allows us to build it with CMake v4
+if [ "$(uname)" = 'Darwin' ]; then
```
xavierchanth, line 8: We can add more checks here if we need to support more platforms with Unix sed

cmake/cjson-patch.sh
```diff
@@ -0,0 +1,14 @@
...
+  # Unix sed
+  sed -e '2s/0/24/' -i '' CMakeLists.txt
+else
+  # GNU sed
+  sed -e '2s/0/24/' -i CMakeLists.txt
```
xavierchanth, line 13: This reads replace the `0` on line 2 with a `24`, and do it in place without a backup file.

xavierchanth: Since the cjson version is pinned, line 2 is always: `cmake_minimum_required(VERSION 3.0)` and becomes `cmake_minimum_required(VERSION 3.24)`

cpswan (approved):

PR #702, opened 2026-08-27
feat: at_activate supports 'proxy:' root server specs (activation via 443)

cconstab:
> [!NOTE]
> **Ready for review.** CI is green (12/12 including memcheck functional tests). Downstream: atsign-foundation/noports#2870 and the csshnpd `atsdk.cmake` pin bump depend on this PR merging.

## What this does

Ports the Dart `AtRootDomain` / `at_lookup` root-server convention (which the Dart `at_activate` via `at_onboarding_cli` already supports) to the C `at_activate`, so atsigns can be activated from networks where only port 443 egress is allowed:

| `-r` value | behavior |
|---|---|
| `host` | ask the atDirectory at `host:64` (unchanged default) |
| `host:port` | ask the atDirectory at `host:port` (**new** — the port was previously hardcoded to 64) |
| `proxy:host` | no atDirectory: the atServer address is `host:64` |
| `proxy:host:port` | no atDirectory: the atServer address is `host:port` |

e.g.

```
at_activate onboard -a @mydevice -c <cram> -r "proxy:proxy0001.atsign.org:443"
at_activate enroll  -a @mydevice -s <otp> ... -r "proxy:proxy0001.atsign.org:443"
```

## Changes

- New `atauth_resolve_atserver` (`packages/atauth/src/resolve_atserver.{c,h}`): parses the root spec and either returns the proxy address directly or delegates to `atclient_utils_find_atserver_address` with the parsed host/port. Host length and port are validated; a bare `proxy:` errors out.
- `run_onboard_command.c` and `run_enroll_command.c` both use the resolver in place of the hardcoded `atclient_utils_find_atserver_address(root_domain, 64, ...)` call. Ownership is unchanged — the resolver's host string is freed by `atclient_authenticate_options_free` exactly like the directory-lookup result was.
- `-r/--rootServer` help text documents the forms (both the onboard and enroll parsers).

## Testing so far

- Build clean.
- Smoke: `onboard -r "proxy:localhost:12399"` logs `Connecting to the atServer via the reverse proxy localhost:12399 (no atDirectory lookup)` and proceeds straight to CRAM (no atDirectory contact); `proxy:` with no host fails with a clear error; regression `-r root.atsign.org` still queries the real atDirectory.

## Remaining end-to-end verification (tracked on the noports `cconstab/c-2831-integration` branch)

- [x] E2E: activate a real atsign through a real reverse proxy on 443 (onboard + enroll paths)
- [x] Regression: activate a real atsign the normal way
- [x] Consider whether `atclient_utils_find_atserver_address` callers elsewhere in the SDK deserve the same convention (e.g. a shared parser in atclient rather than atauth) — kept local to atauth here to keep the change minimal

## Context

Part of making the NoPorts C daemon fully operable on 443-only networks: csshnpd gets the same convention for its runtime connections in atsign-foundation/noports#2870, alongside ESCR/shared-443 relay support in atsign-foundation/noports#2865.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/atauth/src/resolve_atserver.c
+++ b/packages/atauth/src/resolve_atserver.c
@@ -39,11 +39,13 @@ int atauth_resolve_atserver(const char *root_spec, const char *atsign, char **at
     }
     memcpy(host, hostport, host_len);
     host[host_len] = '\0';
-    port = (uint16_t)atoi(colon + 1);
-    if (port == 0) {
+    char *port_end = NULL;
+    long port_val = strtol(colon + 1, &port_end, 10);
+    if (port_end == colon + 1 || *port_end != '\0' || port_val < 1 || port_val > UINT16_MAX) {
       atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Invalid port in root server spec: %s\n", root_spec);
       return 1;
     }
+    port = (uint16_t)port_val;
   } else {
     if (strlen(hostport) > MAX_HOST_LEN) {
       atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Invalid host in root server spec: %s\n", root_spec);
--- a/packages/atauth/src/resolve_atserver.c
+++ b/packages/atauth/src/resolve_atserver.c
@@ -57,10 +57,14 @@ int atauth_resolve_atserver(const char *root_spec, const char *atsign, char **at
     atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_INFO,
                  "Connecting to the atServer via the reverse proxy %s:%u (no atDirectory lookup)\n", host,
                  (unsigned int)port);
-    *atserver_host = strdup(host);
+    // malloc+copy rather than strdup: strdup is not declared under strict
+    // C99 without feature test macros (breaks -Werror builds on glibc)
+    size_t host_size = strlen(host) + 1;
+    *atserver_host = malloc(host_size);
     if (*atserver_host == NULL) {
       return 1;
     }
+    memcpy(*atserver_host, host, host_size);
     *atserver_port = port;
     return 0;
   }
--- a/packages/atauth/src/resolve_atserver.c
+++ b/packages/atauth/src/resolve_atserver.c
@@ -1,0 +1,69 @@
+#include "resolve_atserver.h"
+#include <atclient/atclient_utils.h>
+#include <atlogger/atlogger.h>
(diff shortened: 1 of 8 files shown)


cconstab:
Added a second commit that turned out to be required for proxy mode to work at all: `atclient_connection_connect` was writing a bare `\r\n` after consuming the '@' banner. Against a real atServer that elicits `error:AT0003 invalid verb` on **every** connection (harmless but wrong); against the protocol-aware reverse proxy it is fatal — the proxy routes on the first client line, and an empty line makes it reply directory-style with its own address (no '@') and close. Removing the CRLF exchange (keeping the banner read) fixes both: verified `onboard` via `proxy:proxy0001.atsign.org:443` now completes a full CRAM round trip to the routed atServer, and the direct path behaves identically. This also fixes every atclient consumer (csshnpd etc.) for proxy use — atsign-foundation/noports#2870 depends on it.

packages/atauth/src/resolve_atserver.c
```diff
@@ -0,0 +1,73 @@
...
+  const char *colon = strchr(hostport, ':');
+  if (colon != NULL) {
+    size_t host_len = (size_t)(colon - hostport);
+    if (host_len == 0 || host_len > MAX_HOST_LEN) {
+      atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Invalid host in root server spec: %s\n", root_spec);
+      return 1;
+    }
+    memcpy(host, hostport, host_len);
+    host[host_len] = '\0';
+    port = (uint16_t)atoi(colon + 1);
```
srieteja: This parses an optional :port from --rootServer/-r with atoi() then casts straight to uint16_t. There's no range check — only port == 0 is rejected. So 70000 wraps to 4464 and -1 wraps to 65535, both non-zero, both silently accepted. The wrapped value then gets used as the real connection port, so a malformed -r spec dials a wrong host:port with a generic connection failure instead of the intended "Invalid port in root server spec" error.

cconstab: Good catch — fixed in 1b374c93: port is now parsed with `strtol` and rejected unless it's all digits in 1–65535 (no trailing garbage, no wrap). Verified: `proxy:localhost:70000`, `:-1`, `:443x`, `:0`, and a trailing bare `:` all fail with "Invalid port in root server spec"; `:12399` and boundary `:65535` still connect.

srieteja (approved):

PR #705, opened 2026-08-28
fix: route the enroll connection through 'proxy:' root specs with from:

### LINKED ISSUE #704: at_activate enroll fails through 'proxy:' root server specs — first line must be from:
PR #702 added `proxy:` root server spec support, tested with `at_activate onboard` — which works because CRAM authentication's first line on the connection is `from:`. **`at_activate enroll` through a proxy spec fails**, and the failure mode is subtle.

## Root cause

The protocol-aware reverse proxy (at_services `at_secondary_proxy`, e.g. `proxy0001.atsign.org:443`) routes a connection using the **first line the client sends**, and only understands `from:` (`secondary_connection_bridge.dart` — `if (!command.startsWith("from:"))` → it writes its own address back and closes, emulating an atDirectory answer).

The enroll flow's first command on the fresh unauthenticated connection is `lookup:all:publickey@atsign` (from `atclient_get_public_key`), so the proxy answers `proxy0001.atsign.org:443\n` and closes; the enroll command then fails with "Failed to get the default encryption public key".

Verified against the live proxy:

| first line sent | result |
|---|---|
| `lookup:all:publickey@x` | `proxy0001.atsign.org:443` + EOF (not routed) |
| `from:@x`, then `lookup:all:publickey@x` | routed; lookup returns the public key |

The atServer still accepts unauthenticated `lookup:`/`enroll:` verbs after a bare `from:` (challenge discarded), so the fix is cheap.

## Fix

In `run_enroll_command.c`, when the root spec has the `proxy:` prefix, send a `from:<atsign>\r\n` exchange immediately after `atclient_start_atserver_connection` and discard the response, before the first lookup. Onboard needs no change (CRAM already leads with `from:`), and the post-approval PKAM reconnects are also `from:`-first.

The same fix has been applied to the Arduino snapshot (at_client_arduino).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

cconstab:
## Summary

Fixes #704. `at_activate enroll` through a `proxy:` root server spec failed at its first command:

```
SENT: "lookup:all:publickey@cconstab"
RECV: "proxy0001.atsign.org:443"
ERROR: Failed to get the default encryption public key
```

The protocol-aware reverse proxy routes a connection using the **first line the client sends** and only understands `from:` (`at_secondary_proxy`'s `secondary_connection_bridge.dart`) — anything else gets the proxy's own address back (its atDirectory-emulation behavior) and the socket closed. Onboard works because CRAM leads with `from:`; enroll led with a lookup.

## Fix

When the root spec has the `proxy:` prefix, `atauth_enroll_command` sends a `from:<atsign>` exchange immediately after connecting and discards the challenge; the connection is then bridged and all subsequent commands flow through. Direct (non-proxy) enrollment is unchanged.

## Verification (live proxy, bogus OTP)

Rebuilt `at_activate` with this patch and ran `enroll -r "proxy:proxy0001.atsign.org:443"`:

1. `from:` challenge received through the proxy ✔
2. `lookup:all:publickey@cconstab` returns the real public key ✔
3. `enroll:request` reaches the atServer — rejected only with `AT0022-Exception: invalid otp` (the OTP was deliberately bogus) ✔

The same fix already landed in the Arduino snapshot as atsign-foundation/at_client_arduino#9.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/atauth/src/run_enroll_command.c
+++ b/packages/atauth/src/run_enroll_command.c
@@ -72,6 +72,35 @@ int atauth_enroll_command(const char *atsign, const char *root_domain, const cha
     goto free_auth_options;
   }
 
+  if (strncmp(root_domain, "proxy:", strlen("proxy:")) == 0) {
+    // A protocol-aware reverse proxy (e.g. proxy0001.atsign.org:443) routes a
+    // connection using the FIRST line the client sends, and only understands
+    // 'from:' - any other first line makes it answer with its own address and
+    // close the socket. The enroll flow's first real command is a lookup, so
+    // in proxy mode issue a from: exchange first and discard the challenge;
+    // the connection is then bridged to the atServer and every subsequent
+    // command flows through. (The atServer still accepts unauthenticated
+    // lookup/enroll verbs after a bare from:.)
+    const char *atsign_without_at = (atsign[0] == '@') ? atsign + 1 : atsign;
+    const size_t from_cmd_size = strlen("from:") + strlen(atsign_without_at) + strlen("\r\n") + 1;
+    char *from_cmd = malloc(from_cmd_size);
+    if (from_cmd == NULL) {
+      atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to allocate memory for from_cmd\n");
+      ret = 1;
+      goto free_auth_options;
+    }
+    snprintf(from_cmd, from_cmd_size, "from:%s\r\n", atsign_without_at);
+    unsigned char from_recv[256];
+    size_t from_recv_len = 0;
+    ret = atclient_connection_send(&atclient.atserver_connection, (unsigned char *)from_cmd, strlen(from_cmd),
+                                   from_recv, sizeof(from_recv), &from_recv_len);
+    free(from_cmd);
+    if (ret != 0) {
+      atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to route the connection via the reverse proxy\n");
+      goto free_auth_options;
+    }
+  }
+
   atclient_atkeys atkeys;
   atclient_atkeys_init(&atkeys);
 


cconstab:
Parity note: this mirrors exactly how the Dart client already handles proxies — published `at_onboarding_cli` (1.13.0+) has `_sendFromCommandIfUsingProxy()` in `at_onboarding_service_impl.dart`, which sends a bare unauthenticated `from:<atsign>` before the enroll flow's first command when the root domain has the `proxy:` prefix, and discards the response. Verified live: the Dart at_activate's verbose log shows `enroll: from: command successful` immediately before its `lookup:publickey`, and it reaches the atServer through proxy0001.atsign.org:443 (rejected only for a deliberately bogus OTP, same as this PR's C verification).

JeremyTubongbanua (approved):
**2 nits that are a little contradicting to each other, so i'll just approve!**

1. `atclient_string_utils_atsign_without_at exists`. Declared in `packages/atclient/include/atclient/string_utils.h:71` and implemented in `packages/atclient/src/string_utils.c:123`. It mallocs and returns a new string (caller must free). So instead of:const char *atsign_without_at = (atsign[0] == '@') ? atsign + 1 : atsign;

You should use:

```
char *atsign_without_at = NULL;
ret = atclient_string_utils_atsign_without_at(atsign, &atsign_without_at);
if (ret != 0) {
  atlogger_log(TAG, ATLOGGER_LOGGING_LEVEL_ERROR, "Failed to strip @ from atsign\n");
  goto free_auth_options;
}
// ... use it ...
free(atsign_without_at);
```

2. from_recv buffer size. Yes, you should name it. Looking at how the codebase does it elsewhere (atclient.c:219, atclient.c:450):

const size_t challenge_size = 256;

So the consistent pattern would be something like:

const size_t from_recv_size = 256;
unsigned char from_recv[from_recv_size];
then to describe buffer length, we do `sizeof(unsigned char) * from_recv_size`

